PR #15, opened 2022-02-22
Features/definitions collection proposal.

edgonmsft:
As was being discussed [here](https://github.com/microsoft/dev-container-spec/issues/7).

I wanted to writeup a more formal proposal about the way it could work to foment discussion.


proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
...
+### Definitions
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| type | string | singleContainer/orchestrated |
+| categories | string | The categories where it belongs |
+| architectures | string | Supported architectures |
+| baseOs | string | The operating system in which the definition is based |
+| options | array | List of user selected options at the moment the user selects them in a client application. |
+| styles | array | Variations of the generated definition that contain different information. (e.g. with example code or data.) |
```
Chuxel, line 66: Isn't this what we call "variants" today? Why styles? Right now this just drives the Dockerfile, bit in concept it could be broader which this seems to allude to.  Or is this intended to be more of a "type"?

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Features
+
+| Property | Type | Description |
+|:--- |:--- |:--- |
+| requires | array | List of features that are required for this feature to work. |
```
Chuxel, line 72: This could quickly get into a dependency graph where we'd need to start having the equivalent of lockfiles as well. It will also make it hearder for security minded organizations have their own copy of these that are not public, so we'd need to consider that as well. How would someone create their own private set from a public one?

With that in mind, would it make sense to separate this idea into another proposal instead? I don't know that its explicitly required for what is described here.

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
...
+
+# IDs
+
+TODO
+
+# Search
+
+Search would be implemented by client applications but the definitions themselves should include enough information for this to happen. Besides the typed properties defined above keywords could be:
```
Chuxel, line 80: Perhaps it is worth calling out that some keywords may be defined by clients as having an affect on UX or any "scoring" done for suggestions.

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
...
+| :--- | :--- | :--- |
+| id | string | Id of the feature/definition. The ids should be unique in the contest of the repository where they exist. |
+| displayName | string | Name of the feature/definition |
+| shortDescription | string | Optional one sentence description |
+| description | string | Description of the feature/definition |
+| repositoryPath | string | Path to the repository and folder that contains the root of the code for the feature/definition. |
+| keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
+| version | string | Versioning information for the feature/definition. |
+| files | array | List of links for the download files with checksums and optional signature information. |
+
```
Chuxel, line 56: I mentioned this in the issue, but I'd suggest there should be a `metadata` property here as well for free-form data. Today there's things in these definition files that are specific to MS policies that could land under this category, and its likely other organizations have the same needs, but in different forms. Either that or we probably need to introduce the idea of a "bill of material" to track things that are added. There's existing formats for this already.

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
+# Proposal
+
+The proposal is to arrive to an schema where different communities and organizations can create "collections" of features/definitions that they manage, support and present to users/
+
+This collection data should have enough information so that different clients can search them and present them to users.
+
+For this we propose the existence of two different files:
+- Collection file: (devcontainer-collection.json) Groups of features/definitions that will be handled as the same origin (repository/publisher). Contains the metadata pertinent for each one, checksum and other information.
```
Chuxel, line 8: For my own understanding, "microsoft/vscode-dev-containers" would be a "collection" in this context, correct?

Chuxel (commented):
A few first pass comments:

proposals/devcontainer-collections.md
joshspicer: That's my understanding too.  This is a barebones "example" (which is what we'll turn the successors of vscode-dev-container into): https://github.com/joshspicer/devcontainer

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
...
+- Feature/Definition: (devcontainer-feature.json/devcontainer-definition.json) Information on a particular feature/definition required to add it to a collection. This file is optional and exists to support scenarios where one repository represents a collection and others represent each feature.
```
joshspicer, line 9: I think we'll want to make a decision here, either:

1. Create one top-level  "devcontainer-collection.json` file that has  `features: [ ]` and `definitions: []` arrays storing the metadata.
2. Each individual feature and definition has their own "snippet" of metadata, that the GitHub action will then stitch together

proposals/devcontainer-collections.md
joshspicer: I think we'll want to stick to the current naming convention for `features`, so an `options` object like this:

```json
         "options": {
            "VARIANT": {
              "type": "string",
              "proposals": [
                "1", 
                "1.16", 
                "1.17", 
                "1-bullseye", 
                "1.16-bullseye", 
                "1.17-bullseye", 
                "1-buster", 
                "1.16-buster", 
                "1.17-buster"
              ],
              "default": "1-bullseye"
            },
            "NODE_VERSION": {
              "type": "string",
              "proposals": [
                "none",
                "lts/*",
                "16",
                "14",
                "12",
                "10"
              ],
              "default": "none"
            }
          }
        },
```

proposals/devcontainer-collections.md
joshspicer: I've opened issue https://github.com/microsoft/dev-container-spec/issues/16 on this topic/proposal 👍

proposals/devcontainer-collections.md
Chuxel: Ah - yeah if this is options, We could, however, enable referencing an image directly in this model as well, where an option just updates the "image" field rather than args.  We could have a type of "image" and "buildArg" to distinguish between the two.

proposals/devcontainer-collections.md
joshspicer: Renamed my example to `devcontainer-collection.json` to be consistent!

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
+# Proposal
```
joshspicer, line 1: Am I missing the reference the concept of centralized 'index(es)'.  

RIght now I have this flat file with a single reference to a single collection: https://github.com/joshspicer/devcontainer-index

We would then add pointers here to collections we "endorse".  Others in the community would be able to produce their own indexes, which point whatever set of collections they would like.  Additional indexes could be added to Codespaces via org policy, or via vscode by a setting.

joshspicer:
As more of a "visual" - this is my understanding of the spec for a subset of the vscode-dev-container definitions. I'm looking to merge in the next day or so as an interactive example to this proposal.  
https://github.com/microsoft/vscode-dev-containers/pull/1307

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,110 @@
...
+| files | array | List of links for the download files with checksums and optional signature information. |
+
+### Definitions
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| type | string | singleContainer/orchestrated |
+| categories | string | The categories where it belongs |
+| architectures | string | Supported architectures |
+| baseOs | string | The operating system in which the definition is based |
```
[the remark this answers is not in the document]
edgonmsft: I'm leaning more towards keeping everything in keywords and (like you mention in one of the other comments) keeping a list of recommended and recognized keywords that developers can use. And even a list of recommended 'types' of keywords. I removed 'baseOs' and 'architectures' following this idea.

proposals/devcontainer-collections.md
edgonmsft: Changed to variants, I forgot it already existed. The idea is more of a declaration of different release files that correspond to the same definition. Added some more information.

proposals/devcontainer-collections.md
edgonmsft: Removed and just left a TODO

proposals/devcontainer-collections.md
edgonmsft: Added that we should create a list of well-known keywords including those defined by clients.

proposals/devcontainer-collections.md
edgonmsft: Added the `metadata` property.

proposals/devcontainer-collections.md
edgonmsft: Yeah, that is the idea.

proposals/devcontainer-collections.md
edgonmsft: I see the indexes as more of a client feature, where clients define how to store and search it. Also, how users would add a reference to a collection to search their internal features for example.

proposals/devcontainer-collections.md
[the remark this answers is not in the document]
Chuxel: Do we want to instead document known tags/keywords?  e.g. certainly whether arm64 is support will be of great interest to macOS users. I could see filtering on what works on the chip architecture right off the bat. A list of architecture may be one that does warrant its own property otherwise.

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,114 @@
...
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| type | string | singleContainer/orchestrated |
+| options | array | List of user selected options at the moment the user selects them in a client application. |
+| variants | array | Variations of the generated definition that contain different information. (e.g. with example code or data.) |
+
+Variants:
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| name | string | Name of the variant |
```
chrmarti, line 69: Is each variant a definition by itself? (I guess that would require a link/pointer to that definition?)

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,114 @@
...
+| description | string | Description of the feature/definition |
+| metadata | any | Freeform data added by users for their own purposes. |
+| repositoryPath | string | Path to the repository and folder that contains the root of the code for the feature/definition. |
+| keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
+| version | string | Versioning information for the feature/definition. |
+| files | array | List of links for the download files with checksums and optional signature information. |
+
+### Definitions
+
+| Property | Type | Description |
```
chrmarti, line 60: At a later point we could add an icon (that can be shown in the UI next to the name) and an optional link to a sample.

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,114 @@
...
+# Search
+
+Search would be implemented by client applications but the definitions themselves should include enough information for this to happen. Besides the typed properties defined above keywords could be:
+
+- Purpose. WebApp, WebService, IDE extension, etc.
+- Platform. Node, Ruby on Rails, .Net, etc.
+- Language. TS, JS, C#, Java, etc.
+- Other Apps. Databases, simulators, etc.
+- Dependencies. Bare metal, Azure, AWS, etc.
+- Organization.
```
chrmarti, line 89: Additional data that we might want to represent as separate properties (not as keywords) because we are using them programmatically (currently parsed from the definition's README.md):
- A flag for 'core definitions' to show only these when we don't have any other way to filter down the list of definitions in the UI. (Not sure how to represent that when we accept anyones definitions, maybe this should be a list on the products' side.)
- Lists of platforms/languages to filter down the list of definitions when we know some of the file types used in the repository. (This can be done as keywords if that's preferred.)
- A flag indicating compatibility with Codespaces: To be product-neutral, we might want to have a list of known compatible products and a list of known incompatible products, so each product can then decide whether to include a definition that does not list the product at all.

proposals/devcontainer-collections.md
chrmarti: Is that where we would have the 3 collections from Codespaces, VS Code and the community to start with?

proposals/devcontainer-collections.md
joshspicer: Got it.  For now i've placed all the definitions into the a single file in vscode-dev-container, but could easily stitch that together with the existing GitHub Action

https://github.com/microsoft/vscode-dev-containers/blob/main/devcontainer-collection.json

proposals/devcontainer-collections.md
joshspicer: Yea, I want us to publish an index file that will point to our three collections.  All we need to do is hardcode a single index file, and _that_ will point to the `N` collections we want to pull from by default.

proposals/devcontainer-collections.md
```diff
@@ -0,0 +1,114 @@
...
+| displayName | string | Name of the feature/definition |
+| shortDescription | string | Optional one sentence description |
+| description | string | Description of the feature/definition |
+| metadata | any | Freeform data added by users for their own purposes. |
+| repositoryPath | string | Path to the repository and folder that contains the root of the code for the feature/definition. |
+| keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
+| version | string | Versioning information for the feature/definition. |
+| files | array | List of links for the download files with checksums and optional signature information. |
+
+### Definitions
```
joshspicer, line 58: Do we still want to call this 'definitions' :)

Issue #21, opened 2022-03-24
Feature build performance


chrmarti:
Currently features are built in a single layer of the image, so when a feature is added, removed or its options are modified, all features need to be rebuilt.

Building each feature in a separate layer would partially improve the use of the layer cache, but when the user changes a feature, that would still trigger a rebuild of all the feature layers coming after that.

A more promising approach is to use a multi-stage build and build each feature in a separate stage. One additional advantage is that these stages can be built in parallel. The challenge with it is that a feature needs to build its 'layer' in a separate folder, so in the last build stage we can copy all these feature layers (folders) from the different feature stages. Not all features might be simple to implement in this way, but we can keep supporting the existing way of building features while we work on updating popular features to the new model.

A feature would come with a set of scripts that are run in the following order:
- `validate-prereqs`: Runs as root in a separate build stage using a "build image".
  - This can install build time dependencies.
- `acquire`: Runs as regular user after `validate-prereqs` in the build image.
  - This needs to copy the result to a 'layer' folder that is self-contained in that it can be copied to a different image and run there.
- `configure`: Runs as root in the final image build after the 'layer' from `acquire` is copied over. Based on a "run image".
  - This can run additional configuration or even installation steps. This will run on each rebuild, so doing less here is better.

A new-style feature could omit the `configure` and `validate-prereqs` scripts if it doesn't need them. Old-style features could come only with a `configure` script.

The build image and the run image must be binary compatible in that binaries built for the build image must be able to run in the run image. (E.g., Debian/Ubuntu vs. Alpine. x64 vs. ARM.)

On dependencies: Plain runtime dependencies between features don't affect this approach. When there are build-time dependencies between features, we can declare these on the feature (separate to runtime dependencies) and let the depending feature's stage run with the layer from its dependencies copied in. (This will reduce parallelization of building stages. Unclear if build-time dependencies are common.)

jsma:
Apologies if this is the wrong place to ask, but would the proposal above address the fact that running ``devcontainer build`` followed by ``devcontainer open`` opens VS Code but then in the output one of the first things it does is start building stuff again? 

```
[1690 ms] Start: Run: docker events --format {{json .}} --filter event=start
[1692 ms] Start: Run: docker-compose --project-name myproject -f /Users/username/Projects/myproject/docker-compose.yml build
```
[...]

PR #27, opened 2022-04-18
Proposal for Features functionality and spec.

edgonmsft:
After the beta functionality of features included today and with the feedback we have received so far, here is a proposal to define Features and how the work, get community feedback and create a formalized spec that can be implemented.

### DIFF
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -45,6 +45,8 @@ The properties of the file are as follows:
 | id | string | Id of the feature/definition. The id should be unique in the context of the repository/published package where they exist. |
 | name | string | Name of the feature/definition |
 | description | string | Description of the feature/definition |
+| documentationURL | string | Url that targets the documentation of the feature. |
+| licenseURL | string | Url that points to the license of the feature. |
 | version | string | Version of the feature. |
 | metadata | any | Freeform data added by users for their own purposes. |
 | keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -1,0 +1,141 @@
+# Dev container features reference
+
+Dev container features provide a smooth path for customizing your container definitions. Features are self contained units of code meant to facilitate creation of custom images for application or development containers.
+
+From a practical point of view, features are folders that contain units of code with different entrypoints for different lifecycle events.
+
+Features can be defined by a `feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
+
+Features are to be executed in sequence as defined in `devcontainer.json`.
+
+## Folder Structure 
+
+A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
+
++-- feature
+|    +-- feature.json
+|    +-- validate.sh (default)
+|    +-- aquire.sh (default)
+|    +-- install.sh (default)
+|    +-- (other files)
+
+In case `feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
+
+In case there is intent to create a set of features that share code, it is possible to create a feature collection in the following way:
+
+collectionFolder
++-- feature-collection.json
++-- common (or similar)
+|    +-- (library files)
++-- feature1
+|    +-- feature.json
+|    +-- install.sh (or others)
+|    +-- (other files)
++-- feature2
+(etc)
+
+## feature.json properties
+
+the `feature.json` file defines information about the feature to be used by any tool that helps a user utilize them, and the way the feature will be executed.
+
+The properties of the file are as follows:
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | Id of the feature/definition. The id should be unique in the context of the repository/published package where they exist. |
+| name | string | Name of the feature/definition |
+| description | string | Description of the feature/definition |
+| version | string | Version of the feature. |
+| metadata | any | Freeform data added by users for their own purposes. |
+| keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
+| install.app | string | App to execute.ll (Optional, defaults to /bin/sh) |
+| install.file | string | Parameters/script to execute. (Defaults to validate.sh) |
+| options | object | Possible options to be passed as environment variables to the execution of the scripts |
+| containerEnv | object | A set of name value pairs that sets or overrides environment variables. |
+| privileged | boolean | If preveleged mode is required by the feature. |
+| init | boolean | If its necesarry to run `init`. |
+| capAdd | array | Additional capabilities needed by the feature. |
+| securityOpt | array | Security options needed by the feature. |
+| entrypoint | string | Set if the feature sets requires an entrypoint. |
+
+Options
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | Id of the option |
+| id.type | string | Type of the option |
+| id.enum | string array | possible values |
+| id.default | string | default value for the option |
+| id.description | string | Description for the option |
+
+## feature-collection.json properties.
+
+A feature collection file is a compilation of the `feature.json` files for each individual feature.
+
+If the application finds a `feature-collection.json` file in the root of a downloaded tar file then it uses that file as to find the particular feature that will be executed.
+
+In addition to the list of features included, the file includes the following properties.
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| name | string | Name of the collection |
+| reference | string | Reference information of the repository and path where the code is stored. |
+| version | string | Version of the code. |
+
+In most cases the `feature-collection.json` file can be generated automatically at the moment of creating a release.
+
+## devcontainer.json properties
+
+Features are referenced in `devcontainer.json`  where the `features` tag consists of an array with the ordered list of features to be included in the container image.
+
+The properties are:
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | reference to the particular feature to be included. |
+| options | object | Type of the option |
+
+The `id` is the main reference point for how to find and download a particular feature. `id` can be defined in any of the following ways:
+
+| Type | Description |
+| :--- | :--- |
+| feature  | A simple name references a feature included with the application used to execute `devcontainer.json` |
+| organization/repository/{feature or collectionl/feature}/@version | Reference to a particular release in a repository |
+| https://<../URI/..>/devcontainer-features.tgz#{optional feature} | Direct reference to a file download. |
+| ./{local-path}  -or-  ../{local-path} | A path relative to devcontainer.json where a feature or feature collection can be found. |
+
+## Authoring
+
+Features can be authored in a number of languages, the most straight forward being bash scripts. If a feature is authored in a different language information about it should be included in the metadata so that users can make an informed choice about it.
+
+Reference information about the application required to execute the feature should be included in `feature.json` in the metadata section.
+
+Applications should default to `/bin/sh` for features that do not include this information.
+
+If the feature is included in a folder as part of the repository that contains `devcontainer.json` no other steps are necessary.
+
+## Release
+
+A release is created when the objective is to have other users use a feature.
+
+A release consists of the following:
+
+1.- Tar file with all the included files for the feature or feature collection.
+2.- A copy of the `feature.json` or `feature-collection.json` file that defines the contents of the tar file with additional information added to validate it:
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| version | string | SemVer version of the release. |
+| checksum | string | Checksum of the tar file. |
+
+## Execution
+
+The application that implements features should:
+
+- Features are executed in the order defined in devcontainer.json
+- It should be possible to execute a feature multiple times with different parameters.
+- Features are used to create an image that can be used to create a container or not.
+- Parameters like `privileged`, ``init` are included if just 1 feature requires them.
+- Parameters like `capAdd`, `securityOp`  are concatenated.
+- ContainerEnv is added before the feature is executed and persisted.
+
+TODO
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -12,26 +12,30 @@ Features are to be executed in sequence as defined in `devcontainer.json`.
 
 A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
 
+```
 +-- feature
 |    +-- devcontainer-feature.json
-|    +-- aquire.sh (default)
 |    +-- install.sh (default)
 |    +-- (other files)
+```
 
 In case `devcontainer-feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
 
 In case there is intent to create a set of features that share code, it is possible to create a feature collection in the following way:
 
+```
 collectionFolder
 +-- devcontainer-collection.json
 +-- common (or similar)
 |    +-- (library files)
 +-- feature1
 |    +-- devcontainer-feature.json
-|    +-- install.sh (or others)
+|    +-- install.sh
 |    +-- (other files)
 +-- feature2
 (etc)
+```
+
 
 ## devcontainer-feature.json properties
 
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -4,7 +4,7 @@ Dev container features provide a smooth path for customizing your container defi
 
 From a practical point of view, features are folders that contain units of code with different entrypoints for different lifecycle events.
 
-Features can be defined by a `feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
+Features can be defined by a `devcontainer-feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
 
 Features are to be executed in sequence as defined in `devcontainer.json`.
 
@@ -14,17 +14,16 @@ A feature is a self contained entity in a folder. A feature release would be a t
 
 +-- feature
 |    +-- feature.json
-|    +-- validate.sh (default)
 |    +-- aquire.sh (default)
 |    +-- install.sh (default)
 |    +-- (other files)
 
-In case `feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
+In case `devcontainer-feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
 
 In case there is intent to create a set of features that share code, it is possible to create a feature collection in the following way:
 
 collectionFolder
-+-- feature-collection.json
++-- devcontainer.collection.json
 +-- common (or similar)
 |    +-- (library files)
 +-- feature1
@@ -36,7 +35,7 @@ collectionFolder
 
 ## feature.json properties
 
-the `feature.json` file defines information about the feature to be used by any tool that helps a user utilize them, and the way the feature will be executed.
+the `devcontainer-feature.json` file defines information about the feature to be used by any tool that helps a user utilize them, and the way the feature will be executed.
 
 The properties of the file are as follows:
 
@@ -50,8 +49,8 @@ The properties of the file are as follows:
 | version | string | Version of the feature. |
 | metadata | any | Freeform data added by users for their own purposes. |
 | keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
-| install.app | string | App to execute.ll (Optional, defaults to /bin/sh) |
-| install.file | string | Parameters/script to execute. (Defaults to validate.sh) |
+| install.app | string | App to execute.|
+| install.file | string | Parameters/script to execute |
 | options | object | Possible options to be passed as environment variables to the execution of the scripts |
 | containerEnv | object | A set of name value pairs that sets or overrides environment variables. |
 | privileged | boolean | If preveleged mode is required by the feature. |
@@ -70,11 +69,11 @@ Options
 | id.default | string | default value for the option |
 | id.description | string | Description for the option |
 
-## feature-collection.json properties.
+## devcontainer-collection.json properties.
 
-A feature collection file is a compilation of the `feature.json` files for each individual feature.
+A feature collection file is a compilation of the `devcontainer-feature.json` files for each individual feature. It inlines all the information in each `devcontainer-feature.json` for each feature.
 
-If the application finds a `feature-collection.json` file in the root of a downloaded tar file then it uses that file as to find the particular feature that will be executed.
+If the application finds a `devcontainer-collection.json` file in the root of a downloaded tar file then it uses that file as to find the particular feature that will be executed.
 
 In addition to the list of features included, the file includes the following properties.
 
@@ -84,7 +83,7 @@ In addition to the list of features included, the file includes the following pr
 | reference | string | Reference information of the repository and path where the code is stored. |
 | version | string | Version of the code. |
 
-In most cases the `feature-collection.json` file can be generated automatically at the moment of creating a release.
+In most cases the `devcontainer-collection.json` file can be generated automatically at the moment of creating a release.
 
 ## devcontainer.json properties
 
@@ -109,7 +108,7 @@ The `id` is the main reference point for how to find and download a particular f
 
 Features can be authored in a number of languages, the most straight forward being bash scripts. If a feature is authored in a different language information about it should be included in the metadata so that users can make an informed choice about it.
 
-Reference information about the application required to execute the feature should be included in `feature.json` in the metadata section.
+Reference information about the application required to execute the feature should be included in `devcontainer-feature.json` in the metadata section.
 
 Applications should default to `/bin/sh` for features that do not include this information.
 
@@ -122,7 +121,7 @@ A release is created when the objective is to have other users use a feature.
 A release consists of the following:
 
 1.- Tar file with all the included files for the feature or feature collection.
-2.- A copy of the `feature.json` or `feature-collection.json` file that defines the contents of the tar file with additional information added to validate it:
+2.- A copy of the `devcontainer-feature.json` or `devcontainer-collection.json` file that defines the contents of the tar file with additional information added to validate it:
 
 | Property | Type | Description |
 | :--- | :--- | :--- |
@@ -138,6 +137,6 @@ The application that implements features should:
 - Features are used to create an image that can be used to create a container or not.
 - Parameters like `privileged`, ``init` are included if just 1 feature requires them.
 - Parameters like `capAdd`, `securityOp`  are concatenated.
-- ContainerEnv is added before the feature is executed and persisted.
-
-TODO
+- ContainerEnv is added before the feature is executed as `ENV` commands in the docker file.
+- Features are added to an image in two passes. One for `aquire` scripts and another for `install` scripts.
+- Each script executes as its own layer to aid in caching and rebuilding.
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -13,7 +13,7 @@ Features are to be executed in sequence as defined in `devcontainer.json`.
 A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
 
 +-- feature
-|    +-- feature.json
+|    +-- devcontainer-feature.json
 |    +-- aquire.sh (default)
 |    +-- install.sh (default)
 |    +-- (other files)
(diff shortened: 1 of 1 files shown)


proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+
+The properties of the file are as follows:
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | Id of the feature/definition. The id should be unique in the context of the repository/published package where they exist. |
+| name | string | Name of the feature/definition |
+| description | string | Description of the feature/definition |
+| documentationURL | string | Url that targets the documentation of the feature. |
+| licenseURL | string | Url that points to the license of the feature. |
```
bhack, line 49: Can we add an [OSI boolean](https://opensource.org/licenses)?
I will be faster to filter then regex on license remote files.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+| version | string | Version of the feature. |
+| metadata | any | Freeform data added by users for their own purposes. |
```
chrmarti, line 51: What would that be for example?

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+| keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
+| install.app | string | App to execute.ll (Optional, defaults to /bin/sh) |
```
chrmarti, line 53: This can be covered by the `#!` header of the script file.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+| install.file | string | Parameters/script to execute. (Defaults to validate.sh) |
```
chrmarti, line 54: Would that be needed for all script files? Let's omit this for simplicity, the predefined names do not seem to be a limitation.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
+# Dev container features reference
+
+Dev container features provide a smooth path for customizing your container definitions. Features are self contained units of code meant to facilitate creation of custom images for application or development containers.
+
+From a practical point of view, features are folders that contain units of code with different entrypoints for different lifecycle events.
+
+Features can be defined by a `feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
```
chrmarti, line 7: Let's use a distinctive name to make it easy to associate a schema file (and possibly other editing support) to it. `devcontainer-feature.json` would work well. Similarly, let's use `devcontainer-feature-collection.json` or shorter `devcontainer-features.json`.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+|    +-- aquire.sh (default)
+|    +-- install.sh (default)
+|    +-- (other files)
+
+In case `feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
+
+In case there is intent to create a set of features that share code, it is possible to create a feature collection in the following way:
+
+collectionFolder
++-- feature-collection.json
```
chrmarti, line 27: I liked the idea of being able to distribute definitions/templates and features together. Should we make this a `devcontainer-definitions.json` to contain templates and features?

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+| :--- | :--- | :--- |
+| id | string | Id of the option |
+| id.type | string | Type of the option |
+| id.enum | string array | possible values |
+| id.default | string | default value for the option |
+| id.description | string | Description for the option |
+
+## feature-collection.json properties.
+
+A feature collection file is a compilation of the `feature.json` files for each individual feature.
```
chrmarti, line 75: Does it inline the same information or just reference each feature file?

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+| Type | Description |
+| :--- | :--- |
+| feature  | A simple name references a feature included with the application used to execute `devcontainer.json` |
+| organization/repository/{feature or collectionl/feature}/@version | Reference to a particular release in a repository |
+| https://<../URI/..>/devcontainer-features.tgz#{optional feature} | Direct reference to a file download. |
+| ./{local-path}  -or-  ../{local-path} | A path relative to devcontainer.json where a feature or feature collection can be found. |
+
+## Authoring
+
+Features can be authored in a number of languages, the most straight forward being bash scripts. If a feature is authored in a different language information about it should be included in the metadata so that users can make an informed choice about it.
```
chrmarti, line 110: Alpine doesn't have Bash preinstalled. Let's refer to the Bourne shell (/bin/sh) to maximize compatibility.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+
+A release consists of the following:
+
+1.- Tar file with all the included files for the feature or feature collection.
+2.- A copy of the `feature.json` or `feature-collection.json` file that defines the contents of the tar file with additional information added to validate it:
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| version | string | SemVer version of the release. |
+| checksum | string | Checksum of the tar file. |
```
chrmarti, line 130: Adding the checksum of the tar file to the tar file itself is not possible. Are you considering having two files?

With NPM packages the version goes into the package and the checksum is computed on the package tar. We could add the checksum to the index file we have discussed and make the tar file the only release file.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+Features are to be executed in sequence as defined in `devcontainer.json`.
+
+## Folder Structure 
+
+A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
+
++-- feature
+|    +-- feature.json
+|    +-- validate.sh (default)
+|    +-- aquire.sh (default)
```
chrmarti, line 18: Let's have `aquire.sh` run in a separate build stage to help with caching. (https://github.com/devcontainers/spec/issues/21)

proposals/devcontainer-features.md
joshspicer: I have `devcontainer-collection.json` here: https://github.com/microsoft/vscode-dev-containers/blob/main/devcontainer-collection.json

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| version | string | SemVer version of the release. |
+| checksum | string | Checksum of the tar file. |
+
+## Execution
+
+The application that implements features should:
+
+- Features are executed in the order defined in devcontainer.json
```
chrmarti, line 136: We should keep this declarative for several reasons:
- Usability: When there are implicit dependencies the user has no way knowing what they are and we have no way of changing them over time without risking to break existing configurations.
- Parallelization: We want to preserve the option to parallelize installation in the future, using multi-stage builds or otherwise.
- Explicit dependencies: We want to preserve the option to add explicit dependencies, these would change the order of installation (potentially change it over time as dependencies change).

One example was the creation of a user: Since the user is specified in the devcontainer.json as a top-level option, it makes more sense to deal with that in general instead of as part of a feature (the `common` feature?). It might be easiest if we create the user before installing any features if that user does not exist yet.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,143 @@
...
+- It should be possible to execute a feature multiple times with different parameters.
```
chrmarti, line 137: - Should a feature opt-in to supporting multiple runs? Otherwise this will make writing features harder.
- Is it up to the feature to decide which run will be on the PATH by default?
- How would a feature provide a version switcher to change which install is on the PATH?

proposals/devcontainer-features.md
joshspicer: nit: "acquire"

proposals/devcontainer-features.md
edgonmsft: I liked the shorthand of just `feature.json` but if we agree on that change, I can make it.

proposals/devcontainer-features.md
edgonmsft: With other discussions we haven't seen much use with `validate` so we can define them as that features are added in to passes, first all `aquire` scripts are executed and then all `install` ones.

proposals/devcontainer-features.md
edgonmsft: Changed to `devcontainer-collection` to signify that its a group.

proposals/devcontainer-features.md
edgonmsft: I'm thinking mostly used for showing on UX that show features in a search experience that is not contemplated in the definition and wanted to reserve the word.

proposals/devcontainer-features.md
edgonmsft: Not sure I get what you mean.

proposals/devcontainer-features.md
edgonmsft: Inlines he information so that its the only file needed to show information to the user.

proposals/devcontainer-features.md
edgonmsft: Agreed, I removed the reference to bash.

proposals/devcontainer-features.md
edgonmsft: I think I wasn't clear, but the idea is that the file is included in the tar file so that during processing we have the information. We only add the checksum to the file that gets posted as a release next to it so that the download can be validated. I think it just simplifies things this way.

proposals/devcontainer-features.md
edgonmsft: - I agree that features can opt-in to supporting multiple runs, perhaps as a propertie of `devcontianer-feature.json` that we could validate in some way.
- While making changes to the implementation of the current features I added parameters to control it and with true. That reinforces the idea that features need to declare if they support multiple runs or not.
- I feel like that is more a feature of a tool that a user would use at runtime.

proposals/devcontainer-features.md
edgonmsft: - I like the idea of having/using the cached features for things like users, although I'm not sure if defaulting to running it before other features is the best idea.
I feel like there are two ways of dealing with this. One is the array format which makes it clear to the user in what order things are going to run and the user has total control of that. Not sure I like the idea of things changing from under me if there is an update somewhere upstream. Declaring dependencies could actually be done in `devcontainer.json` instead of as part of the feature itself. If we do that the tool can more easily parallelize, but it complicates the user experience in that now they have to declare the dependencies instead of just using an in order execution.

Thoughts? @Chuxel @jkeech @joshspicer

proposals/devcontainer-features.md
Chuxel: Might be worth calling out what is intended to be in these scripts.  If acquire can run as non-root and outputs to a folder, then something needs to be able to install native dependencies intended to do the acquisition... and the acquire script can't do it.  If we're not making that distinction right now, then one entry point may be clearer TBH.

The interesting thing here is the buildpack implementation would have root-requiring steps run *first* and then steps that can land just in a folder second... and nothing after.  I think that's the reverse of what is here. The reason the issue @chrmarti referenced included both a "verify-prereqs.sh" and a "configure.sh" was that you needed to be able to install acquisition dependencies -- e.g. curl or wget -- before you can execute a script.    That has to run before "aquire.sh".  So it was verify-prereqs => acquire => configure. The buildpack model then doesn't have "configure" at all... it more configure => acquire

edgonmsft: Ok I see I get what you mean. I think this gets into a interesting discussion about exactly how much work we expect/want a feature to do. On one side there is the do everything required to get something to work, on the other is just install what is asked and expect the user to put other features for covering those things. I think the idea is somewhere in between but it gets interesting because I feel like the current features do too many things to not depend in something else. Which is good if they are intended to be completely independent but complicates using multiple features together.

Chuxel: The dependencies I'm talking about are those needed by the aquire script to run. e.g. if you're running a build, you need gcc, if you're just downloading something you need curl/wget, if you're cloning something you need git. There's no way someone using a feature can know that up front without reading docs... and we're trying to prevent them from having to do that.

proposals/devcontainer-features.md
chrmarti: To clarify: Dependencies would be declared by features themselves (not in the devcontainer.json), the user should not have to know about these. We don't have dependencies at the moment, but since we have already investigated adding them, I use them as one example to show that having `"features"` in the devcontainer.json determine an execution order will constrain us in the future. (Even more important is the usability argument above.)

On the user example: The user is configured with the top-level `"remoteUser"` independently of features. That suggests that we should address it also independently of features. Since features then depend on that user, we would create it before we run the feature install scripts. (Having a feature create the `"remoteUser"` doesn't seem to make sense since that user needs to exist also without any features for the configuration to be valid.)

Another argument for keeping "features" declarative is that it will make adding auto-detection of features in the future simpler. (If order matters, we need logic to arrange that for the detected features.)

proposals/devcontainer-features.md
Chuxel: One possible way to resolve this if we want to keep two scripts is to reverse the order.

install => acquire

Install focuses on adding anything that requires root access including native OS capabilities, config, updating the home directory, etc.  acquire then focuses on basic acquisition and has a more flexible caching model. 

We should also be sure to call out and set an environment variable for a target path for acquire... for caching to work we'd need to dictate where the install happens.

proposals/devcontainer-features.md
bhack: The scope was related to having something machine readable so that we can quckly filter also in the UI.

How can I quickly filter OSS only elements?

proposals/devcontainer-features.md
chrmarti: We can always use `install.sh` (assuming `validate.sh` is a typo), there is no need to make this configurable.

proposals/devcontainer-features.md
chrmarti: Does that make it sound like it is a collection of dev containers. Would `devcontainer-definitions.json` go as the abbreviation for `devcontainer-templates-and-features.json`?

proposals/devcontainer-features.md
chrmarti: It would be easier to let a feature add options for installing multiple versions (maybe with some additional support for doing that). That would allow the feature to control installation of all requested versions in a single run instead of having it run multiple times. With that approach IntelliSense in the devcontainer.json would likely include a hint if a feature supports multiple versions (e.g., an `"additionalVersions"` property).

proposals/devcontainer-features.md
edgonmsft: Bein configurable is so that features can be created in other languages, if its not present we do fallback to `install.sh`

proposals/devcontainer-features.md
edgonmsft: I like the idea of `devcontainer-collection.json` since that way the file can contain both templates and features related to `devcontainers` as users know them.

proposals/devcontainer-features.md
edgonmsft: Ok I get what you mean @Chuxel, yeah my idea is that each feature is responsible of validating its os dependencies and installing them. I also like the idea of splitting them for caching purposes.

@chrmarti does that mean that the first script would be expected to run parallel but the second script would run in sequence for when a feature depends on a different one?

Chuxel: > To optimize for ease of use for feature authors, we could call the final script run as root in the "run image" install.sh. E.g., a simple feature might just run apt-get install ... in that script.

>
>acquire.sh would run as root in the "build image" and prepare a folder that is later available in the install.sh. E.g., a simple performance-aware feature might just download binaries to that folder and later unpack them to system folders in install.sh. (This allows the download to run in parallel with other acquire.sh scripts.) (https://github.com/devcontainers/spec/issues/21 also has validate-prereqs.sh run as root in the build image and acquire.sh run as regular user in the build image, but I'm not sure we need this distinction.)

>
>To fully optimize performance a feature would do everything in acquire.sh (for parallelization and cachability) and use install.sh only to add symlinks to its files in the final image.

@chrmarti Yeah, though I don't think we should assume `acqure.sh` will have root access in all implementations. It should place content in a specific location that is passed into it that already has privs set. In some ways it would be ideal if it never had root access so it discourages adding content anywhere outside of the folders that will be cached/copied. (There just may not always be a non-root user to do that - so maybe we don't want to force it?)

I think we do need to ensure proper sequencing here however. To describe the intent, those phases were:

1. [Root]: Prepare the (build and run) image with any native OS dependencies - and you could in concept install external .debs here too.
2. [Non-Root]: Acquire, perform local setup/build, and place the result in a prescribed location
3. [Root, Optional for many things] Do any final configuration. This should be very minimal and often optional.  e.g. run ldconfig

Perhaps we should decouple this conversation from the initial spec for features.  Feels like we may want to have a meeting on these nuances.  This can add just `install.sh`, and we decide what the other scripts are for #21 separately.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
+
++-- feature
+|    +-- feature.json
+|    +-- aquire.sh (default)
+|    +-- install.sh (default)
+|    +-- (other files)
+
+In case `devcontainer-feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
```
bamurtaugh, line 21:
```suggestion
In case `devcontainer-feature.json`  does not include a reference for the lifecycle scripts, the application will look for the default script names and will execute them if available.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
++-- feature1
+|    +-- feature.json
+|    +-- install.sh (or others)
+|    +-- (other files)
++-- feature2
+(etc)
+
+## feature.json properties
+
+the `devcontainer-feature.json` file defines information about the feature to be used by any tool that helps a user utilize them, and the way the feature will be executed.
```
bamurtaugh, line 38:
```suggestion
the `devcontainer-feature.json` file defines information about the feature to be used by any supporting tools and the way the feature will be executed.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+Features can be defined by a `devcontainer-feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
+
+Features are to be executed in sequence as defined in `devcontainer.json`.
+
+## Folder Structure 
+
+A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
+
++-- feature
+|    +-- feature.json
```
bamurtaugh, line 16: Should this be `devcontainer-feature.json` instead? If not, I wouldn't be clear at this point of the doc what the difference is between the two files.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+In case `devcontainer-feature.json`  does not include a reference for the lifecycle scripts the application will look for the default script names and will execute them if available.
+
+In case there is intent to create a set of features that share code, it is possible to create a feature collection in the following way:
+
+collectionFolder
++-- devcontainer.collection.json
++-- common (or similar)
+|    +-- (library files)
++-- feature1
+|    +-- feature.json
```
bamurtaugh, line 30: Again confused about `devcontainer-feature.json` vs `feature.json`.

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+|    +-- install.sh (or others)
+|    +-- (other files)
++-- feature2
+(etc)
+
+## feature.json properties
```
bamurtaugh, line 36: Should this be `devcontainer-feature.json`?

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+the `devcontainer-feature.json` file defines information about the feature to be used by any tool that helps a user utilize them, and the way the feature will be executed.
+
+The properties of the file are as follows:
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | Id of the feature/definition. The id should be unique in the context of the repository/published package where they exist. |
```
bamurtaugh, line 44:
```suggestion
| id | string | Id of the feature/definition. The id should be unique in the context of the repository/published package where the feature exists. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| name | string | Name of the feature/definition |
```
bamurtaugh, line 45: Nit: trying to be consistent with periods (either add them to every statement, or remove them from all).
```suggestion
| name | string | Name of the feature/definition. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| description | string | Description of the feature/definition |
```
bamurtaugh, line 46:
```suggestion
| description | string | Description of the feature/definition. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| documentationURL | string | Url that targets the documentation of the feature. |
```
bamurtaugh, line 47: To be consistent with the `licenseURL` definition.

```suggestion
| documentationURL | string | Url that points to the documentation of the feature. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| licenseURL | string | Url that points to the license of the feature. |
+| version | string | Version of the feature. |
+| metadata | any | Freeform data added by users for their own purposes. |
+| keywords | array | List of keywords relevant to a user that would search for this definition/feature. |
+| install.app | string | App to execute.|
+| install.file | string | Parameters/script to execute |
```
bamurtaugh, line 53:
```suggestion
| install.file | string | Parameters/script to execute. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| options | object | Possible options to be passed as environment variables to the execution of the scripts |
```
bamurtaugh, line 54:
```suggestion
| options | object | Possible options to be passed as environment variables to the execution of the scripts. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| containerEnv | object | A set of name value pairs that sets or overrides environment variables. |
+| privileged | boolean | If preveleged mode is required by the feature. |
```
bamurtaugh, line 56:
```suggestion
| privileged | boolean | If privileged mode is required by the feature. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| init | boolean | If its necesarry to run `init`. |
```
bamurtaugh, line 57:
```suggestion
| init | boolean | If it's necessary to run `init`. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| capAdd | array | Additional capabilities needed by the feature. |
+| securityOpt | array | Security options needed by the feature. |
+| entrypoint | string | Set if the feature sets requires an entrypoint. |
```
bamurtaugh, line 60:
```suggestion
| entrypoint | string | Set if the feature requires an entrypoint. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+Options
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | Id of the option |
+| id.type | string | Type of the option |
+| id.enum | string array | possible values |
```
bamurtaugh, line 68:
```suggestion
| id.enum | string array | Possible values |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| id.default | string | default value for the option |
```
bamurtaugh, line 69:
```suggestion
| id.default | string | Default value for the option |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| id.description | string | Description for the option |
+
+## devcontainer-collection.json properties.
```
bamurtaugh, line 72:
```suggestion
## devcontainer-collection.json properties
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+A feature collection file is a compilation of the `devcontainer-feature.json` files for each individual feature. It inlines all the information in each `devcontainer-feature.json` for each feature.
+
+If the application finds a `devcontainer-collection.json` file in the root of a downloaded tar file then it uses that file as to find the particular feature that will be executed.
```
bamurtaugh, line 76:
```suggestion
If the application finds a `devcontainer-collection.json` file in the root of a downloaded tar file, then it uses that file to find the particular feature that will be executed.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+In addition to the list of features included, the file includes the following properties.
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| name | string | Name of the collection |
```
bamurtaugh, line 82:
```suggestion
| name | string | Name of the collection. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| reference | string | Reference information of the repository and path where the code is stored. |
+| version | string | Version of the code. |
+
+In most cases the `devcontainer-collection.json` file can be generated automatically at the moment of creating a release.
```
bamurtaugh, line 86:
```suggestion
In most cases, the `devcontainer-collection.json` file can be generated automatically at the moment of creating a release.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+## devcontainer.json properties
+
+Features are referenced in `devcontainer.json`  where the `features` tag consists of an array with the ordered list of features to be included in the container image.
```
bamurtaugh, line 90:
```suggestion
Features are referenced in `devcontainer.json`, where the `features` tag consists of an array with the ordered list of features to be included in the container image.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+The properties are:
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| id | string | reference to the particular feature to be included. |
```
bamurtaugh, line 95:
```suggestion
| id | string | Reference to the particular feature to be included |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| options | object | Type of the option |
+
+The `id` is the main reference point for how to find and download a particular feature. `id` can be defined in any of the following ways:
+
+| Type | Description |
+| :--- | :--- |
+| feature  | A simple name references a feature included with the application used to execute `devcontainer.json` |
```
bamurtaugh, line 102:
```suggestion
| feature  | A simple name referencing a feature; it's included with the application used to execute `devcontainer.json`. |
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+| organization/repository/{feature or collectionl/feature}/@version | Reference to a particular release in a repository |
+| https://<../URI/..>/devcontainer-features.tgz#{optional feature} | Direct reference to a file download. |
+| ./{local-path}  -or-  ../{local-path} | A path relative to devcontainer.json where a feature or feature collection can be found. |
+
+## Authoring
+
+Features can be authored in a number of languages, the most straight forward being bash scripts. If a feature is authored in a different language information about it should be included in the metadata so that users can make an informed choice about it.
```
bamurtaugh, line 109:
```suggestion
Features can be authored in a number of languages, the most straightforward being bash scripts. If a feature is authored in a different language, information about it should be included in the metadata so that users can make an informed choice about it.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+Reference information about the application required to execute the feature should be included in `devcontainer-feature.json` in the metadata section.
+
+Applications should default to `/bin/sh` for features that do not include this information.
+
+If the feature is included in a folder as part of the repository that contains `devcontainer.json` no other steps are necessary.
```
bamurtaugh, line 115:
```suggestion
If the feature is included in a folder as part of the repository that contains `devcontainer.json`, no other steps are necessary.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Release
+
+A release is created when the objective is to have other users use a feature.
+
+A release consists of the following:
+
+1.- Tar file with all the included files for the feature or feature collection.
+2.- A copy of the `devcontainer-feature.json` or `devcontainer-collection.json` file that defines the contents of the tar file with additional information added to validate it:
```
bamurtaugh, line 124:
```suggestion
1. Tar file with all the included files for the feature or feature collection.
2. A copy of the `devcontainer-feature.json` or `devcontainer-collection.json` file that defines the contents of the tar file, with additional information added to validate it.
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| version | string | SemVer version of the release. |
+| checksum | string | Checksum of the tar file. |
+
+## Execution
+
+The application that implements features should:
```
bamurtaugh, line 133:
```suggestion
There are several things to keep in mind for an application that implements features:
```

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Features are executed in the order defined in devcontainer.json
+- It should be possible to execute a feature multiple times with different parameters.
+- Features are used to create an image that can be used to create a container or not.
+- Parameters like `privileged`, ``init` are included if just 1 feature requires them.
```
bamurtaugh, line 138:
```suggestion
- Parameters like `privileged`, `init` are included if just 1 feature requires them.
```

bamurtaugh (commented):
Thanks for sharing this again!

proposals/devcontainer-features.md
chrmarti: > @chrmarti Yeah, though I don't think we should assume `acqure.sh` will have root access in all implementations. It should place content in a specific location that is passed into it that already has privs set. In some ways it could be ideal if it never had root access so it discourages adding content anywhere outside of the folders that will be cached/copied. (There just may not always be a non-root user to do that - so maybe we don't want to force it?)

Any other implementation would have the ability to run scripts as root when building the image. Having a single script run on the build image will avoid limitations imposed by running one with and one without root access.

> I think we do need to ensure proper sequencing here however. To describe the intent, those phases were:
> 
> 1. [Root]: Prepare the (build and run) image with any native OS dependencies - and you could in concept install external .debs here too.
> 2. [Non-Root]: Acquire, perform local setup/build, and place the result in a prescribed location
> 3. [Root, Optional for many things] Do any final configuration. This should be very minimal and often optional.  e.g. setup and run ldconfig.
> 
> So, from a concurrency perspective, we could do (1) for all features, then concurrently run (2) since it would be writing to separate folders in each case, and finally (3).

Running `acquire.sh` in parallel and then `install.sh` would be a simpler model to get started with.

> Perhaps we should decouple this conversation from the initial spec for features. Feels like we may want to have a meeting on these nuances. This can add just `install.sh`, and we decide what the other scripts are for #21 separately maybe?

Sounds good.

chrmarti: > @chrmarti does that mean that the first script would be expected to run parallel but the second script would run in sequence for when a feature depends on a different one?

That would make sense.

proposals/devcontainer-features.md
chrmarti: I would avoid adding non-essential properties to the spec for now and this can be covered by having an `install.sh` forward the call.

proposals/devcontainer-features.md
Chuxel: > Any other implementation would have the ability to run scripts as root when building the image. Having a single script run on the build image will avoid limitations imposed by running one with and one without root access.
>
> ..
>
> Running acquire.sh in parallel and then install.sh would be a simpler model to get started with.

Yeah, unfortunately this would preclude `acquire.sh` from being part of a build step in a buildpack. It runs as UID 1000 (typically a user called `cnb`) in the build container. Anyway, sounds like we're all agreeing that for now just having `install.sh` makes sense while we iron these other aspects out separately.

proposals/devcontainer-features.md
edgonmsft: Made the changes. cc @joshspicer, @samruddhikhandale

edgonmsft:
I think this is ok as an initial version, thoughts? @Chuxel, @bamurtaugh, @joshspicer, @samruddhikhandale , @chrmarti

proposals/devcontainer-features.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Folder Structure 
+
+A feature is a self contained entity in a folder. A feature release would be a tar file that contains all files part of the feature.
+
++-- feature
+|    +-- devcontainer-feature.json
+|    +-- aquire.sh (default)
+|    +-- install.sh (default)
+|    +-- (other files)
```
Chuxel, line 19: Didn't we discuss removing `acquire.sh` for now? (Its still in this list but misspelled as aquire.sh)

Also, this doesn't format right because it's not a code block in the markdown... everything ends up on the same line.

Chuxel:
Yeah I think as a proposal we could get it in and then have another PR with updates  if that's ok w/@chrmarti and @bamurtaugh  as well. One formatting note above 👆

bamurtaugh:
Sounds good to me! This plus https://github.com/devcontainers/spec/pull/33 can serve as great initial versions of the spec, which we can update in subsequent PRs as we make changes and get community feedback.

proposals/devcontainer-features.md
edgonmsft: Yeah I don't want to block features from receiving multiple values and using them to loop, the idea is that we should also not disallow multiple runs.

edgonmsft:
Merging as a proposal and we can do additional PR's for more changes before adding it to the main spec.

proposals/devcontainer-features.md
bhack: @edgonmsft Do I need to open a new ticket to track this?

edgonmsft: Yep, lets create a new one. I think its a good idea the we should flesh out a little bit more. Thanks!

proposals/devcontainer-features.md
edgonmsft: The main idea of making this changes is to allow users to control the order of execution, especially since we are not supporting dependencies yet and the changes to the OS applied by features might clash between themselves in unpredictable ways that a user might need to fix in this way. 

Thoughts @Chuxel, @bamurtaugh, @jkeech, @joshspicer ?

bamurtaugh: I'm not sure if this would work, but is there a way a user could set a property if the order of their features matters? I think the concerns @chrmarti mentions make sense, but I can see the value in what @edgonmsft mentions about controlling the order of execution, so could we default to declarative but allow setting a property that ordering matters, in case users would like to override this declarative default?

chrmarti: The order that works best will depend on how the features are implemented and could change over time. The user does not have any insight into this. When we add dependencies we want to be in control of the order.

We should ensure the order of installation is always the same (e.g., sort by id), so the outcome is reproducible. If we have any implicit dependencies today, I suggest we take a closer look and find a way to deal with them short term.

joshspicer: My biggest concern with obscuring the installation order of features (which are implemented as  shell scripts with no standardization on _how_ to author them), is that is will significantly hinder the ability of a community member to author their own set of features, and prevent the wider adoption and ecosystem of features. 

We have not landed on a simple, complete mechanism for automated versioning/dependency management despite a fair bit of time and effort researching and discussing options.      We've investigated building on top of existing package managers like `npm`. Our biggest detractor for such an idea is that we aren't just installing the _feature_, but the _feature with the provided options_.  Our "blessed" set of features today already function quite differently given provided options - We would need _strict_ authoring rules for the community, as well as a strictly defined set of metadata and listing of dependencies, to make the declarative features object feasible.  We are talking about reimplementing a package manager like 'apt'.

I think that is too complicated given the problem 'features' is trying to solve (An easy-to-use mechanism to get tool X available in your dev environment, that can auto-update on rebuilds if you want it to)

To the three arguments for implicit ordering:

> Usability: When there are implicit dependencies the user has no way knowing what they are and we have no way of changing them over time without risking to break existing configurations.

I think the opposite is true.  We provide the user with total control over dependencies (we hope there are few with features, but maybe there are 1 or 2).  I think _hiding_ the dependencies will cause more breaks for users.

> Parallelization: We want to preserve the option to parallelize installation in the future, using multi-stage builds or otherwise.

This argument makes sense, we lose a bit of control over installation order if we aren't able to decide that ourselves. 

> Explicit dependencies: We want to preserve the option to add explicit dependencies, these would change the order of installation (potentially change it over time as dependencies change).

This is what we are seeking with the array syntax, an option for users to add explicit dependencies.  If a dependency changes significantly, it is on the author to create a new major version of the feature and for users to make the conscious decision on how they would like to proceed in their dev containers.  I have already implemented a way to pin to a GitHub release or tgz URL, ranges are a next logical step.

joshspicer: Additionally, we'll also need to take a dependency on the base image, which could be based on a yet another base image, both installing tools.  While I think, in a perfect world, we could write tooling to figure everything out for a user, pragmatically I don't think we want to get into that business :D

proposals/devcontainer-features.md
9090y: > Contributor

Fleinsopp (commented):
File change

Fleinsopp:
File change

PR #38, opened 2022-05-27
Add hostRequirements gpus description

anpaskin:
Add description for new hostRequirements gpus option, added by this PR: https://github.com/github/github/pull/222154.

Note: That second diff at the end of the file doesn't actually show up as anything in the file preview, I'm not sure why it's showing up like that.

### DIFF
--- a/docs/specs/devcontainerjson-reference.md
+++ b/docs/specs/devcontainerjson-reference.md
@@ -79,6 +79,7 @@ While `devcontainer.json` does not focus on hardware or VM provisioning, it can 
 | Property | Type | Description |
 |----------|------|-------------|
 | `hostRequirements.cpus` | integer | Indicates the minimum required number of CPUs / virtual CPUs / cores. For example: `"hostRequirements": {"cpus": 2}` |
+| `hostRequirements.gpus` | integer | Indicates the minimum required number of GPU cores. For example: `"hostRequirements": {"cpus": 2}` |
 | `hostRequirements.memory` | string |  A string indicating minimum memory requirements with a `tb`, `gb`, `mb`, or `kb` suffix. For example, `"hostRequirements": {"memory": "4gb"}` |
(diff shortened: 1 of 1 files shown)


docs/specs/devcontainerjson-reference.md
```diff
@@ -79,6 +79,7 @@ While `devcontainer.json` does not focus on hardware or VM provisioning, it can
 | Property | Type | Description |
 |----------|------|-------------|
 | `hostRequirements.cpus` | integer | Indicates the minimum required number of CPUs / virtual CPUs / cores. For example: `"hostRequirements": {"cpus": 2}` |
+| `hostRequirements.gpus` | integer | Indicates the minimum required number of GPU cores. For example: `"hostRequirements": {"cpus": 2}` |
```
jkeech, line 82: The example has a typo (gpus instead of cpus):
```suggestion
| `hostRequirements.gpus` | integer | Indicates the minimum required number of GPU cores. For example: `"hostRequirements": {"gpus": 2}` |
```

jkeech (commented):
Thanks for making this spec proposal!

What does a "GPU core" mean here? Is there perhaps a different unit that would be helpful? It would be great to see some example projects that need GPUs in the dev environment in order to make sure the GPU `hostRequirement` is sufficient to capture the project requirements.

For instance, do projects need certain types of GPUs? Should GPU support be more of a boolean or an enum?

anpaskin:
"GPUs" is probably a better way to express the unit. That's how [this doc](https://docs.microsoft.com/en-us/azure/virtual-machines/ncv3-series) on our GPU SKU words the unit, for example.

As for integer vs. boolean vs. enum, maybe boolean would be more appropriate here. I'm guessing in most cases users would use 1 for the minimum GPUs, just to specify that the SKU for this repo needs to be a GPU SKU. @lostintangent may be able to provide some more context, as well as some example projects that would require GPUs in the dev environment.

Chuxel:
I moved this to an issue which is where we would normally discuss proposals before making a PR: https://github.com/devcontainers/spec/issues/82

jkeech:
I'm going to close this PR for now and we can continue discussion on the issue. Once the design is agreed in the issue, we can revisit the PR to update the spec. Thanks!

PR #39, opened 2022-06-01
Fixes and additions to the spec.

edgonmsft:
cc @bamurtaugh, @Chuxel

### DIFF
--- a/docs/specs/devcontainer-reference.md
+++ b/docs/specs/devcontainer-reference.md
@@ -81,11 +81,17 @@ Mounts allow containers to have access to the underlying machine, share data bet
 
 It is important to note that these mounts are from the underlying compute environment and thus cloud environments might not have access to the same data as a local machine.
 
-## Workspace folder
+A default mount is included so that the source code is accessible from inside the container. Inside the container this mount defaults to `/workspace`.
+`
+## worspaceFolder and worspaceMount
 
-The `workspace-folder` is used for the purpose of identifying the path where the configuration files are found. This path is also automatically included in the mounted folders for the container. This mount is by default created pointing to `/workspace` but can be modified with the [`workspaceMount` and `workspaceFolder`](devcontainerjson-reference.md#image-or-dockerfile-specific-properties) properties.
+The default mount point for the source code can be set with the `workspaceMount` property. This folder should point to the root of a repository (where the `.git` folder is found) so that source control operations work correctly inside the container. By default this value mirrors `workspaceFolder`. 
 
 > **Note**: It's important that this is not considered in the case of [Docker Compose](#docker-compose-based).
+
+The `workspaceFolder` can then be set to a subfolder if it is desired that IDEs should open this folder by default. This value should be taken into account in all scenarios.
+
+See [`workspaceMount` and `workspaceFolder`](devcontainerjson-reference.md#image-or-dockerfile-specific-properties) for reference.
 
 ## Users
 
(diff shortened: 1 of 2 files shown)


docs/specs/devcontainer-reference.md
```diff
@@ -81,13 +81,19 @@ Mounts allow containers to have access to the underlying machine, share data bet
 
 It is important to note that these mounts are from the underlying compute environment and thus cloud environments might not have access to the same data as a local machine.
 
+## Workspace folder
+
+The `workspace-folder` is used for the purpose of identifying the path where the configuration files are found. This path is also automatically included in the mounted folders for the container. This mount is by default created pointing to `/workspace` but can be modified with the [`workspaceMount` and `workspaceFolder`](devcontainerjson-reference.md#image-or-dockerfile-specific-properties) properties.
```
bamurtaugh, line 86:
```suggestion
The `workspaceFolder` is used for the purpose of identifying the path where the configuration files are found. This path is also automatically included in the mounted folders for the container. This mount is by default created pointing to `/workspace` but can be modified with the [`workspaceMount` and `workspaceFolder`](devcontainerjson-reference.md#image-or-dockerfile-specific-properties) properties.
```

docs/specs/devcontainer-reference.md
```diff
@@ -81,13 +81,19 @@ Mounts allow containers to have access to the underlying machine, share data bet
...
+
+> [!NOTE]
+> Its important that this is not considered in the case of [Docker compose](#docker-compose-based).
```
bamurtaugh, line 89:
```suggestion
> **Note**: It's important that this is not considered in the case of [Docker Compose](#docker-compose-based).
```

docs/specs/devcontainer-reference.md
```diff
@@ -81,13 +81,19 @@ Mounts allow containers to have access to the underlying machine, share data bet
...
+
 ## Users
 
 Users control the permissions of applications executed in the containers, allowing the developer to control them. The specification takes into account two types of user definitions:
 
-* Container User: The user that will be used for all operations that run inside a container. It is used to run lifecycle commands among others.
-* Remote User: In case the developer wants to use different users for different purposes, the remote user is the user that connecting tools should use to execute operations inside the container.
-
+* Container User: The user that will be used for all operations that run inside a container. This allows the ENTRYPOINT for the image to execute with different permissions than the developer.
+* Remote User: Is used to run the [lifecycle](#lifecycle) scripts insde the container. Also this is the user that tools and editors that connect to the container should use to run their processes.
```
bamurtaugh, line 96:
```suggestion
* Remote User: Used to run the [lifecycle](#lifecycle) scripts inside the container. This is also the user tools and editors that connect to the container should use to run their processes.
```

docs/specs/supporting-tools.md
```diff
@@ -67,6 +67,8 @@ A [codespace](https://docs.github.com/en/codespaces/overview) is a development e
 
 > **Tip:** If you've already built a codespace and connected to it, be sure to run **Codespaces: Rebuild Container** from the Command Palette (`kbstyle(F1)`) to pick up any changes you make.
 
+> **Tip** Codespaces does implement an auto `workspace-folder` mount in **Docker compose** scenarios.
```
bamurtaugh, line 70:
```suggestion
> **Tip** Codespaces implements an auto `workspaceFolder` mount in **Docker Compose** scenarios.
```

bamurtaugh (commented):
Thanks for making these updates! Including them in the site: https://github.com/devcontainers/devcontainers.github.io/pull/6.

docs/specs/devcontainer-reference.md
Chuxel: I think we'd want to explain the relationship of the mount point and the workspace folder a bit more particularly for Remote - Containers.

The workspaceMount should point at to the root of a repository to ensure that git operations work, even if you are actually trying to open a sub-folder in the repository.  You can then set workspaceFolder to a sub-folder under the mount point in the container if you want to open a sub-folder by default.

I think it's also worth pointing out that the `workspaceMount` property is only used in the orchestrator-less scenario since orchestrators have their own ways to mount content (e.g. Docker Compose).

edgonmsft: Applied this comments.

docs/specs/devcontainer-reference.md
```diff
@@ -140,13 +146,14 @@ This step executes the following:
...
-- If the `wait-for` property is defined, then execution should stop at the specified property.
+- At the end of the container creation step, a set of commands are executed inside the **main** container: `onCreateCommand`, `updateContentCommand` and `postCreateCommand`. This set of commands is executed in sequence on a container the first time it's created and depending on the creation parameters received. You can learn more in the [documentation on lifecycle scripts](devcontainerjson-reference.md#lifecycle-scripts). By default, `postCreateCommand` is executed in the background after reporting the successful creation of the development environment.
+- If the `waitFor` property is defined, then execution should stop at the specified property.
+- `userEnvProbe` is used to define the way environment variables are read from the container before executing the lifecycle hooks.
```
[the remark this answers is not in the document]
edgonmsft: Made the changes.

PR #46, opened 2022-06-20
Fix incorrect property name

Chuxel:
Noticed one spot that still says `workspace-folder` instead of `workspaceFolder`.

@bamurtaugh web content also seems to have this error in it.


bamurtaugh:
Thanks @Chuxel! Fixed in containers.dev in https://github.com/devcontainers/spec/pull/46.

Issue #43, opened 2022-06-16
Feature installation order


chrmarti:
As we are preparing to open up features to community contributions, we found that some features need to install after other features. This issue proposes to add support for soft-dependencies to arrive at a reproducible ordering following these ordering requirements:

A feature can declare a list of other features that should be installed first if they are configured to be installed. (The soft-dependency itself does not add another feature to the list of features to be installed.)

E.g., the NodeJS feature would list the Python feature because it has to install a system version of Python if no version of Python is already installed:
```jsonc
{
    "id": "node",
    "installAfter": [ "devcontainers/features/python" ],
    // ...
}
```

Features that an install in any order will be installed in a fixed order (e.g., ordered by id) to ensure the result is always the same.

The "features" map in the devcontainer.json will remain a map. See https://github.com/devcontainers/spec/pull/27/files#r869018039 for a discussion.

To allow for progress on user contributable features, we can use a stop-gap measure like an ordered list of features that depend on a specific installation order and either bake that into the CLI at build time or fetch it at runtime. This list will be temporary and documented as such.

/cc @Chuxel @jkeech @edgonmsft @joshspicer @bamurtaugh @alexdima

Chuxel:
@chrmarti We'll still need some way to allow users ordering explicit in the feature object in devcontainer.json.

One way to do that could be to allow "installAfter" to also be explicitly mentioned by users in the object syntax as well.  e.g.

```
"features": {
    "common": "latest",
    "docker-in-docker": {
        "installAfter": "common"
    }
}
```

Explicit ordering allows end users to:
1.	Have a well-known and predictable outcome as they add more features (given features need to adapt to the image contents). Reproducible image builds is a key best practice for creating container images.  An unordered dependency graph is not transparent or predictable over time since execution order influences what each feature does. Adding a new feature to the end of the list will not affect anything above it, while a random order can result in more material changes.
    - Examples include the relationship between the Node and Python features, and the Git from Source feature and most other features. Order can result in different behaviors in installing OS dependencies.
    - Similar problems can occur between features authored by multiple users that have no knowledge of one another – and thus no possible dependency graph.
2.	Have a way for a developer to easily add a local script that runs as a part of this sequence, without getting into dependency management or complex conditional logic that adapts to different ordering or image scenarios. 
    - Existing examples include installing the preview dev CLI into the Azure CLI, installing more go tools (via go install), adding desktop customization for the desktop-lite script, etc. 
[...]

Issue #1, opened 2022-01-18
Move settings and extensions properties to vscode namespace


chrmarti:
These VS Code-specific properties are currently top-level. The proposal is to introduce a `vscode` property for VS Code-specific configuration and move these two properties there.

We can copy the properties in the schema, keeping the old properties as 'deprecated' for compatibility.

One open question is how we can include runtime-specific parts of the JSON schema. For those runtimes we know of, we can reference or inline the schema. For now I suggest to keep these properties in the same schema file.

Interestingly settings and extensions also apply to Codespaces which already has a `codespaces` property for its own configuration.

/cc @bamurtaugh @Chuxel @jkeech

chrmarti:
Doing this in 2 stages will allow the support for the new properties to roll out before they become used more widely:
- [ ] February
  - [ ] Add "vscode" property with "extensions" and "settings" subproperties to implementation and schema.
  - [ ] Show a warning when the old and the new property (both "extensions" or both "settings") are used in the same devcontainer.json.
- [ ] March
  - [ ] Update definitions in vscode-dev-containers to use the new properties.
  - [ ] Prefer the new properties when editing by UI command (only 2 cases).

At a future point we might want offer a code action to move the old properties over to the new properties (e.g., when there are more of them). No need to break existing configurations though.

For the devcontainer-features.json we can deprecate the old properties in the schema (but still read them) and start using the new ones.

"customizations" I propose to defer until we have an actual user/customer for it.

Chuxel:
@chrmarti Given our intent to have this be an open spec, I think including a `customizations` property will be important. We do not want adoption of the format to be slowed by a perception that there is no place for other tools while they experiment or add provisional support for the format.

bamurtaugh:
> I worry about name conflicts if it's a "free for all" at the top level. Perhaps we should describe this in contributing.md?

Documenting `customizations` in `contributing.md` makes sense to me. I assume we'd want to wait to describe it until it's actually implemented (not just an fyi right now, so that users don't try to use it already). Unless we wanted to give a more forward look and mention it right now (perhaps linking to this issue; folks who get to this repo may be interested in `How can I contribute my own subproperties`), then update `contributing.md` once it's actually implemented?

Chuxel:
Yeah, I was thinking once we've finalized this proposal.

chrmarti:
Adding `customizations` to the check list above. 👍

[...]

PR #48, opened 2022-06-21
Updating features proposal after latest discussions.

edgonmsft:
@Chuxel, @chrmarti, @alexdima, @jkeech, @joshspicer, @samruddhikhandale 

closes https://github.com/devcontainers/spec/issues/43

### DIFF
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -6,7 +6,19 @@ From a practical point of view, features are folders that contain units of code 
 
 Features can be defined by a `devcontainer-feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
 
-Features are to be executed in sequence as defined in `devcontainer.json`.
+By default features are executed by an order selected by the tool that implements the spec based in the following:
+
+- `runsAfter` property defined as part of `devcontainer-feature.json`.
+- id.
+
+The tool uses the `runsAfter` propery to intelligently manage this order and ensure that if there are relationships betwen the features they are respected.
+
+In case an end user needs to control the order of execution we are adding the `overrideFeatureInstallOrder` to `devcontainer.json`
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| overrideFeatureInstallOrder | array | Array made of the Id's of the features in the order the user wants them to be installed. |
+
 
 ## Folder Structure 
 
@@ -90,12 +102,11 @@ In most cases, the `devcontainer-collection.json` file can be generated automati
 
 ## devcontainer.json properties
 
-Features are referenced in `devcontainer.json` , where the `features` tag consists of an array with the ordered list of features to be included in the container image.
+Features are referenced in `devcontainer.json` , where the `features` tag consists of an object tag starting with the id of the feature. and including the values of the options to pass to the feature itself.
 
 The properties are:
 | Property | Type | Description |
 | :--- | :--- | :--- |
-| id | string | Reference to the particular feature to be included. |
 | options | object | Type of the option .|
 
 The `id` is the main reference point for how to find and download a particular feature. `id` can be defined in any of the following ways:
@@ -135,11 +146,9 @@ A release consists of the following:
 
 There are several things to keep in mind for an application that implements features:
 
-- Features are executed in the order defined in devcontainer.json
-- It should be possible to execute a feature multiple times with different parameters.
+- The order of execution of features is determined by the application based on the `installAfter` property by feature authors and can be overriden by users if necesarry with the `overrideFeatureInstallOrder` in `devcontainer.json`.
 - Features are used to create an image that can be used to create a container or not.
 - Parameters like `privileged`, `init` are included if just 1 feature requires them.
 - Parameters like `capAdd`, `securityOp`  are concatenated.
 - ContainerEnv is added before the feature is executed as `ENV` commands in the docker file.
-- Features are added to an image in two passes. One for `aquire` scripts and another for `install` scripts.
-- Each script executes as its own layer to aid in caching and rebuilding.
+- Each feature script executes as its own layer to aid in caching and rebuilding.
--- a/proposals/devcontainer-features.md
+++ b/proposals/devcontainer-features.md
@@ -10,7 +10,7 @@ By default, features are installed in an order selected by the implementing tool
 
 If any of the following properties are provided in the feature's `devcontainer-feature.json`, or the user's `devcontainer.json`, the order indicated by the propert(ies) are respected.
 
-- `runsAfter` property defined as part of `devcontainer-feature.json`.
+- `installsAfter` property defined as part of `devcontainer-feature.json`.
 - `id`.
 
 The tool uses the `runsAfter` property to intelligently manage this order and ensure that if there are relationships between the features, they are respected.
@@ -79,6 +79,8 @@ The properties of the file are as follows:
 | capAdd | array | Additional capabilities needed by the feature. |
 | securityOpt | array | Security options needed by the feature. |
 | entrypoint | string | Set if the feature requires an entrypoint. |
+| customizations | object | Product specific properties, each namespace under `customizations` is treated as a separate set of properties. For each of this sets the object is parsed, values are replaced while arrays are set as a union. |
+| installsAfter | array | Array of Id's of features that should execute before this one. Allows control for feature authors on soft dependencies between different features. |
 
 Options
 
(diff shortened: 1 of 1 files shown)


proposals/devcontainer-features.md
```diff
@@ -6,7 +6,19 @@ From a practical point of view, features are folders that contain units of code
 
 Features can be defined by a `devcontainer-feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
 
-Features are to be executed in sequence as defined in `devcontainer.json`.
+By default features are executed by an order selected by the tool that implements the spec based in the following:
```
joshspicer, line 9:
```suggestion
By default features are executed by an order selected by the tool that implements the spec based on the following:
```

edgonmsft: Awesome thanks!

proposals/devcontainer-features.md
```diff
@@ -6,7 +6,19 @@ From a practical point of view, features are folders that contain units of code
...
+
+- `runsAfter` property defined as part of `devcontainer-feature.json`.
+- id.
+
+The tool uses the `runsAfter` propery to intelligently manage this order and ensure that if there are relationships betwen the features they are respected.
+
+In case an end user needs to control the order of execution we are adding the `overrideFeatureInstallOrder` to `devcontainer.json`
```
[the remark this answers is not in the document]
edgonmsft: 👍

proposals/devcontainer-features.md
```diff
@@ -90,12 +102,11 @@ In most cases, the `devcontainer-collection.json` file can be generated automati
 
 ## devcontainer.json properties
 
-Features are referenced in `devcontainer.json` , where the `features` tag consists of an array with the ordered list of features to be included in the container image.
+Features are referenced in `devcontainer.json` , where the `features` tag consists of an object tag starting with the id of the feature. and including the values of the options to pass to the feature itself.
```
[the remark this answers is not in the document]
edgonmsft: 👍

proposals/devcontainer-features.md
```diff
@@ -6,7 +6,25 @@ From a practical point of view, features are folders that contain units of code
 
 Features can be defined by a `devcontainer-feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
 
-Features are to be executed in sequence as defined in `devcontainer.json`.
+By default, features are installed in an order selected by the implementing tool. 
+
+If any of the following properties in the feature's `devcontainer-feature.json, or the user's `devcontainer.json` are provided, the ordered indicated by the propert(ies) are respected.
```
bamurtaugh, line 11:
```suggestion
If any of the following properties are provided in the feature's `devcontainer-feature.json`, or the user's `devcontainer.json`, the order indicated by the propert(ies) are respected.
```

proposals/devcontainer-features.md
```diff
@@ -6,7 +6,25 @@ From a practical point of view, features are folders that contain units of code
...
+
+- `runsAfter` property defined as part of `devcontainer-feature.json`.
+- id.
```
bamurtaugh, line 14:
```suggestion
- `id`.
```

proposals/devcontainer-features.md
```diff
@@ -6,7 +6,25 @@ From a practical point of view, features are folders that contain units of code
...
+
+The tool uses the `runsAfter` propery to intelligently manage this order and ensure that if there are relationships betwen the features they are respected.
```
bamurtaugh, line 16:
```suggestion
The tool uses the `runsAfter` property to intelligently manage this order and ensure that if there are relationships between the features, they are respected.
```

proposals/devcontainer-features.md
```diff
@@ -6,7 +6,25 @@ From a practical point of view, features are folders that contain units of code
...
 Features can be defined by a `devcontainer-feature.json` file in the root folder of the feature. The file is optional for backwards compatibility but it is required for any new features being authored.
 
-Features are to be executed in sequence as defined in `devcontainer.json`.
+By default, features are installed in an order selected by the implementing tool. 
+
+If any of the following properties in the feature's `devcontainer-feature.json, or the user's `devcontainer.json` are provided, the ordered indicated by the propert(ies) are respected.
+
+- `runsAfter` property defined as part of `devcontainer-feature.json`.
+- id.
+
```
bamurtaugh, line 15: Should this bulleted list include `overrideFeatureInstallOrder`? Does it need to mention `id`?

proposals/devcontainer-features.md
```diff
@@ -135,11 +152,9 @@ A release consists of the following:
 
 There are several things to keep in mind for an application that implements features:
 
-- Features are executed in the order defined in devcontainer.json
-- It should be possible to execute a feature multiple times with different parameters.
+- The order of execution of features is determined by the application based on the `installAfter` property by feature authors and can be overriden by users if necesarry with the `overrideFeatureInstallOrder` in `devcontainer.json`.
```
bamurtaugh, line 155:
```suggestion
- The order of execution of features is determined by the application, based on the `installAfter` property used by feature authors. It can be overridden by users if necessary with the `overrideFeatureInstallOrder` in `devcontainer.json`.
```

proposals/devcontainer-features.md
```diff
@@ -135,11 +152,9 @@ A release consists of the following:
...
 - Features are used to create an image that can be used to create a container or not.
```
bamurtaugh, line 156: What is the "or not" use case?

proposals/devcontainer-features.md
```diff
@@ -135,11 +152,9 @@ A release consists of the following:
...
 - Parameters like `privileged`, `init` are included if just 1 feature requires them.
```
bamurtaugh, line 157:
```suggestion
- Parameters like `privileged` and `init` are included if just 1 feature requires them.
```

proposals/devcontainer-features.md
```diff
@@ -135,11 +152,9 @@ A release consists of the following:
...
 - Parameters like `capAdd`, `securityOp`  are concatenated.
 - ContainerEnv is added before the feature is executed as `ENV` commands in the docker file.
```
bamurtaugh, line 159: Should this be `containerEnv`?

```suggestion
- `containerEnv` is added before the feature is executed as `ENV` commands in the Dockerfile.
```

proposals/devcontainer-features.md
```diff
@@ -90,13 +108,12 @@ In most cases, the `devcontainer-collection.json` file can be generated automati
...
 
-Features are referenced in `devcontainer.json` , where the `features` tag consists of an array with the ordered list of features to be included in the container image.
+Features are referenced in `devcontainer.json` , where the `features` tag consists of an object tag starting with the id of the feature and including the values of the options to pass to the feature itself.
 
 The properties are:
 | Property | Type | Description |
 | :--- | :--- | :--- |
-| id | string | Reference to the particular feature to be included. |
-| options | object | Type of the option .|
+| options | object | Type of the option. |
```
chrmarti, line 116: We don't need a separate property for options.

chrmarti (commented):
Can we move install.app and install.file to a separate proposal? I think this requires more discussion. (Previously discussed in https://github.com/devcontainers/spec/pull/27#pullrequestreview-961566955.)

proposals/devcontainer-features.md
edgonmsft: I just meant that the code could be executed to just create an image for other purposes.

Chuxel:
Already mentioned elsewhere, but adding here for tracking.  I think we should call out `customizations` being in the schema and the expected behavior for merging properties under it.  (For example, each "namespace" under `customizations` is treated as a separate set of properties. For each set, the object is parsed and leaf values are replaced while leaf arrays are a union).

I wouldn't want the spec to have to update as new tools come in, for example. Right now there's "extensions" and "settings" at the wrong level per #1.

Chuxel:
@edgonmsft @chrmarti One other thought.  I think at one point we'd discussed the idea that a feature could declare the operating systems and architectures it supports.  Given the number of hiccups we've seen with arm64 installs with existing features and PRs like https://github.com/microsoft/vscode-dev-containers/pull/1513 driving things into different base OS distros, I think it may be a good idea to include this in the spec from the beginning.

Didn't we have that included at one point?

We could have normalized architecture values and then support for /etc/os-release detection based on ID or ID_LIKE for distribution.  We can also detect the version in a similar way. Most of the features have some code in them to detect and block based on these values. Moving it into the core spec just makes things easier for feature authors and we can provide UX hints if a feature isn't expected to work given the current environment.

e.g., something like this might work

```
"architectures": [ "x86_64", "arm64"],
"os": "linux",
"version": {
   "debian": ["10", "11"],
   "ubuntu": ["20.04", "18.04"],
   "alpine": true
}
```

...where "true" means any version.

proposals/devcontainer-features.md
edgonmsft: Removed.

edgonmsft:
Also: https://github.com/devcontainers/spec/issues/59

113ll3 (commented):
Review

proposals/devcontainer-features.md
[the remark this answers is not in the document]
Amrfawzy9: ****

PR #72, opened 2022-08-04
clarifications to the feature distribution spec

joshspicer:
Clarifications to the feature spec and feature distribution spec (related to https://github.com/devcontainers/cli/pull/104)


proposals/devcontainer-features.md
```diff
@@ -108,12 +108,14 @@ The `id` format specified dicates how a supporting tool will locate and download
...
-| `https://<uri-to-feature-tgz>` | Direct HTTPS URI to a tarball. | https://github.com/user/repo/releases/devcontainer-feature-go.tgz |
-| `./<path-to-feature-dir>`| A relative directory to folder containing a devcontainer-feature.json. | ./myGoFeature |
+| `https://<..uri..>/devcontainer-feature-<feature>.tgz` | Direct HTTPS URI to a tarball. | https://github.com/user/repo/releases/devcontainer-feature-go.tgz |
+| `./<path-to-feature-dir>`| A relative directory(**) to folder containing a devcontainer-feature.json. | ./myGoFeature |
+
 
-`
 (*) OCI registry must implement the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec).  Some implementors can be [found here](https://oras.land/implementors/).
 
+(**) The provided path is relative to the workspace folder root (the `--workspace-folder` ).
```
samruddhikhandale, line 117: Suggestion: 
(**) The provided path must be inside a directory or subdirectory of the workspace folder root (the `--workspace-folder`).

It can be just me but this sentence sounded a little contradicting to :point-down as the user can have the directory relative to the workspace folder but outside of it. 

> A feature can be referenced directly in a user's [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) by relative path _inside_ the project root directory, or a subdirectory of the project root. Note that the project root directory is indicated by the required `--workspace-folder <...>` parameter.  
> A local feature may not be referenced outside of the project directory (`../` is not allowed), nor is an absolute path allowed.

bamurtaugh (approved):
LGTM! Opened a PR to update containers.dev too: https://github.com/devcontainers/devcontainers.github.io/pull/19.

proposals/devcontainer-features.md
joshspicer: Hmm, I mean for them to sound the same, only the workspace-folder and anything inside of that directory.

proposals/devcontainer-features-distribution.md
```diff
@@ -131,6 +131,8 @@ A feature can be referenced directly in a user's [`devcontainer.json`](/docs/spe
 
 To aid in feature authorship, or in instances where a feature should not be published externally, individual features can be referenced locally from the project's file tree.
 
-A feature can be referenced directly in a user's [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) by relative path _inside_ the project directory.  A local feature may not be referenced outside of the project directory (`../` is not allowed), nor is an absolute path allowed.  
+A feature can be referenced directly in a user's [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) by relative path _inside_ the project root directory, or a subdirectory of the project root. Note that the project root directory is indicated by the required `--workspace-folder <...>` parameter.
```
jkeech, line 134: `--workspace-folder` is not part of the specification. It's a command line flag that our reference implementation of the CLI uses, but another implementor could use a different mechanism to set the root. Can we use language that is anchored in the specification and not in a specific implementation of the specification?

proposals/devcontainer-features-distribution.md
```diff
@@ -131,6 +131,8 @@ A feature can be referenced directly in a user's [`devcontainer.json`](/docs/spe
 
 To aid in feature authorship, or in instances where a feature should not be published externally, individual features can be referenced locally from the project's file tree.
 
-A feature can be referenced directly in a user's [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) by relative path _inside_ the project directory.  A local feature may not be referenced outside of the project directory (`../` is not allowed), nor is an absolute path allowed.  
+A local feature is referenced in a user's [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) by relative path. A local feature is always referenced relative to the project root.
```
chrmarti, line 134: Other paths in the devcontainer.json (e.g., "dockerfile", "dockerComposeFile", "context) are relative to the devcontainer.json's folder. Let's do the same here for consistency.

proposals/devcontainer-features-distribution.md
```diff
@@ -131,6 +131,8 @@ A feature can be referenced directly in a user's [`devcontainer.json`](/docs/spe
...
+
+A local feature may not be referenced outside of the project's root directory (`../` is not allowed), nor is an absolute path allowed.
```
chrmarti, line 136: With my above comment `../` would be allowed as long as it does not go outside the project's root folder.

chrmarti (dismissed):
Great that we clarify this here. I have left a request for aligning with other paths.

proposals/devcontainer-features-distribution.md
jkeech: Can we define the project root as well? Is that based on the repository root, which implies that all devcontainers need to be inside a git repo, or is there some other definition of project root?

I'd rather be more restrictive to start and then open up the set of valid paths later on if needed based on feedback. It's much easier to broaden the set of allowed paths over time than to narrow the set due to backwards compatibility requirements.

proposals/devcontainer-features-distribution.md
joshspicer: So that I understand the behavior:

In @samruddhikhandale 's case with the devcontainer/images repo (that has a [`.devcontainer.json` in the `workspaceFolder`](https://github.com/devcontainers/images/tree/main/src/php)) this change will result in the same path being calculated (so no change needed on her side).

For example, if she had a `.devcontainer/devcontainer.json` in the `workspaceFolder` [for this image](https://github.com/devcontainers/images/tree/main/src/php), then all local features would need to be prefixed with `../`.  

@chrmarti, how do you normally pass the information on where the devcontainer lives downstream in the reference implementation?

proposals/devcontainer-features-distribution.md
chrmarti: The workspace folder (i.e., project folder) is passed to the CLI with `--workspace-folder` and that is used to look up the devcontainer.json (which can be overridden by another CLI option).

I agree on being restrictive. Would we be fine with `.devcontainer/devcontainer.json` not being able to refer to a local feature outside of the `.devcontainer` folder?

proposals/devcontainer-features-distribution.md
chrmarti: @joshspicer Correct. Not sure what you mean by downstream, but the config object has a `configFilePath` pointing at the file (currently using a URI for historical reasons).

proposals/devcontainer-features-distribution.md
joshspicer: I think restricting to the `.devcontainer` folder is good to start.  @samruddhikhandale - the `devcontainers/images` repo would need to adapt to this, right?

samruddhikhandale: > @samruddhikhandale - the devcontainers/images repo would need to adapt to this, right?

Yes!

alexdima (approved):

PR #77, opened 2022-08-08
typo with devcontainer-collection oci media type

joshspicer:
This is the media type the spiked GitHub action implements, and how https://github.com/devcontainers/features/pkgs/container/features is published today.

image

### DIFF
--- a/proposals/devcontainer-features-distribution.md
+++ b/proposals/devcontainer-features-distribution.md
@@ -120,7 +120,7 @@ NAMESPACE=devcontainers/features
 
 oras push ${REGISTRY}/${NAMESPACE}:latest \
         --manifest-config /dev/null:application/vnd.devcontainers \
-                            ./devcontainer-collection.json.:application/vnd.devcontainers.layer.v1+json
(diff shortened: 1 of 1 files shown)


proposals/devcontainer-features-distribution.md
```diff
@@ -120,7 +120,7 @@ NAMESPACE=devcontainers/features
 
 oras push ${REGISTRY}/${NAMESPACE}:latest \
         --manifest-config /dev/null:application/vnd.devcontainers \
-                            ./devcontainer-collection.json.:application/vnd.devcontainers.layer.v1+json
+                            ./devcontainer-collection.json:application/vnd.devcontainers.collection.layer.v1+json
```
samruddhikhandale, line 123: Curious, looking at the `Manifest` in [here](https://github.com/devcontainers/features/pkgs/container/features/32903232?tag=latest), where does the `collection.layer` part come from?

joshspicer: That github rendering is not actually accurate to what I see if I inspect it manually with a tool like `skopeo`

chrmarti (approved):

Issue #83, opened 2022-08-15
Terminal post-attach commands


joshaber:
## Problem

We want to codify the post-attach commands which clients should run in parallel, interactive terminals.

For example: I always want my server (`npm start`) and my file watcher (`npm run watch`) running when I connect.

## Proposal

- The client will decide how to run these commands interactively in a manner appropriate for it.
    - For example, rich clients like VS Code and JetBrains could run them as Terminals, command line clients could ignore them, run them in a tmux session, etc.
- The client will ensure the commands are running each time they attach and recreate them if they are not. 
    - Because this is client-specific, connecting with multiple clients could result in the commands running multiple times.

## Details

1. Modify `postAttachCommand` to:
    ```typescript
    postAttachCommand: string | array | {[key: string}: TerminalPostAttachCommand};

    type TerminalPostAttachCommand = string | {command: string | array, terminal: boolean}
    ```
2. The object form of `postAttachCommand` will default to `terminal: true`.
3. The key is the identifier for the post attach command. This can be referenced by client-specific settings, see below.
4. The value or `command` is the command to run.

Example:
```jsonc
{
  "postAttachCommand": {
    "server": "npm start", // defaults to `terminal: true`
    "watcher": "npm run watch",
    "greetings": {
      "command": "echo hello",
      "terminal": false // The current headless post attach behavior
    }
  },

  // Client-specific customizations using the keys above:
  "customizations": {
    "vscode": {
      "terminals": {
        "server": { // Key matches the `postAttachCommand` key above
          "icon": "bug"
        },
        "watcher": {
          "name": "All-seeing eye",
          "icon": "eye",
          "color": "red"
        }
      }
    }
  }
}
```

## Open Questions
1. Should `terminal: true` be the default behavior for `postAttachCommand`?
    - This would be a breaking change, but looking at [`postAttachCommand` usage in the wild](https://github.com/search?q=%22postAttachCommand%22%3A&type=code) it looks like many of the use cases would either (1) be improved by running in interactive terminals, or (2) surface to the user that `postAttachCommand` is the wrong lifecycle hook for the behavior they want.

cc @Chuxel @jkeech since this is coming out of conversations with y'all

Chuxel:
> `interactive: true` commands will not be run by devcontainer CLI.

[...]

PR #88, opened 2022-09-01
Client-side post-attach commands

joshaber:
Formalize https://github.com/devcontainers/spec/issues/83 into a proposal.

[**Rendered**](https://github.com/devcontainers/spec/blob/joshaber/post-attach-proposal/proposals/client-post-attach-commands.md)

### DIFF
--- a/proposals/client-post-attach-commands.md
+++ b/proposals/client-post-attach-commands.md
@@ -1,0 +1,58 @@
+# Client post-attach commands
+
+`postAttachCommand` is currently a single command which implementations run as part of the container lifecycle.
+
+This proposal makes two changes:
+1. Redefines post-attach commands to be a client concern, separate from container lifecycle.
+2. Allows multiple, named commands in `object` form. This is to both support multiple post-attach commands and to support using the name as an identifier for client-specific customization.
+
+## Motivation
+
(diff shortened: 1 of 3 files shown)


proposals/client-post-attach-commands.md
```diff
@@ -0,0 +1,62 @@
...
+    "watch": "npm run watch",
+    "db": "mysql -u root -p"
+  }
+}
+```
+
+## Spec changes
+
+- `postAttachCommand` will support an `object` form where the key is the unique name of the command and the value is the `string` or `array` command. Duplicate names are unsupport.
+- Implementations will no longer run `postAttachCommand` as part of the container lifecycle scripts but instead defer that responsibility to attaching clients. Clients may run the scripts at any point after they are attached and may be run concurrently with container lifecycle scripts.
```
joshaber, line 28: This is a breaking change to the current behavior, do we need a CLI flag to opt in/out of it?

proposals/client-post-attach-commands.md
```diff
@@ -0,0 +1,62 @@
...
+- Implementations will provide a `read-post-attach-commands` command which will parse the post-attach commands from the config, interpolate variables, and output the final commands. Clients will use this to determine the commands they should run after attaching.
```
joshaber, line 29: Bike shedding on this name welcome!

joshaber:
Replaced by https://github.com/devcontainers/spec/pull/89.

PR #89, opened 2022-09-02
Parallel lifecycle script execution proposal

joshaber:
Coming out of discussions on https://github.com/devcontainers/spec/issues/83.

Propose supporting parallel lifecycle script execution when scripts are provided as objects.


joshaber:
Should I update the schemas in this PR or do that separately?

chrmarti (approved):

joshaber:
As per our conversation earlier today I'm going to go ahead and land this proposal and then follow up with the implementation and finally the schema/docs updates.

Issue #18, opened 2022-03-21
Dev container metadata in image labels


Chuxel:
Unlike some developer centric formats, a non-goal for devcontainer.json is to become yet another multi-container orchestrator format. Instead, its goal is to enable development in containers regardless of how they are orchestrated. With this in mind, allowing image labels be used to set devcontainer.json properties could have significant advantages. In particular, there is a problem with images and devcontainer.json files needing to be paired but are deployed and managed separately, which can lead to disconnects. This is important in scenarios like container runtime settings required to enable debugging (e.g. cap_add: SYS_PTRACE), but applies broadly to any settings the images was purpose built to support.

In particular:

1. When the dev container CLI is used to build the image, these labels can added automatically.
1. Labels can be added to images using a Dockerfile, command line, an orchestrator format, or another image build system like Buildpacks.

This second point also makes it easier to use the metadata when some other mechanism is used to build the container images - whether an orchestrator or another non-Docker CLI like `pack` for Buildpacks.

The json-based nature of devcontainer.json would main this fairly straight forward. To simplify just reusing json blobs in automation scenarios, we can also support a whole structure being added:

```Dockerfile
LABEL dev.containers.metadata='{ "userEnvProbe": "loginInteractiveShell", "customizations": { "vscode": { "extensions": ["ms-python.python", "ms-toolsai.jupyter" ] } } }'
```

For hand editing, common and less complex properties could be referenced directly. 

```Dockerfile
LABEL dev.containers.metadata.userEnvProbe="loginInteractiveShell"
LABEL dev.containers.metadata.customizations.vscode.extensions= '["ms-python.python","ms-toolsai.jupyter"]'
```

More complex `any` type properties could then be encoded json.

```Dockerfile
LABEL dev.containers.metadata.customizations.vscode.settings="{\"some.setting\": \"some-value\"}"
```

In the common case, these labels would be automatically added by the dev container CLI to the image when it is pre-built. (`devcontainer build ...`), but manual entry would enable these additions to be embedded in orchestrator formats as well.

```yaml
services:
  app:
    build:
      context: .
      dockerfile: Dockerfile
      labels:
        - "dev.containers.metadata.userEnvProbe=loginInteractiveShell"
        - "dev.containers.metadata.customizations.vscode.extensions=[\"ms-python.python\",\"ms-toolsai.jupyter\"]"
```

[...]

Issue #60, opened 2022-06-30
Lifecycle hooks support for dev container features


Chuxel:
This proposal is an enhancement idea to the current [dev container features proposal](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md).

---

Currently the proposed dev container features concept can update the contents of a container image and provide needed dev container configuration metadata to use those contents. The proposal includes an `entrypoint` property and this will run as non-interactive process with the default user for the container and is intended for things that should happen at the moment the container starts (like `ENTRYPOINT` in a Dockerfile) - this allows things like user modification that wouldn't be possible later.

However, there are situations where being able to affect something at a later lifecycle stage also makes sense. [Lifecycle scripts](https://containers.dev/implementors/json_reference/#lifecycle-scripts) run as the `remoteUser` and have `remoteEnv` and `userEnvProbe` processing applied to them. This allows commands like `npm install` to execute in the context of the correct user to ensure permissions and other factors are correctly handled.

I'd propose we make these properties available to features as well. For example:
1. Private features (e.g., when a feature is authored by someone in a devops / plat ops team) can enable sharing of standards and best practices within a team that can update over time without having to do a mass devcontainer.json update. Having the ability to use commands like postCreateCommand to fire something one time (e.g. registering a container with a management serrvice) would be very useful.
2. For scenarios like data science, it can be useful to provide features that encompass robust setup behaviors. For example, an analogous feature to "[repo-to-docker](https://github.com/jupyterhub/repo2docker)" could be constructed for a one-stop-shop in getting a fully functional environment.

Similar to what is described in #18 and #19, we would want firing these commands to be baked into the image in some way in the case of pre-building such that you wouldn't need to reference the feature in devcontainer.json to get the benefits.  Assuming we move forward with #18, that could be enough with the idea that the feature would copy any needed scripts to a well known location.

//cc @edgonmsft @jkeech @joshspicer @lostintangent @chrmarti @bamurtaugh

jkeech:
To make sure I understand the proposal, is the idea that feature can contribute lifecycle hook additions? Say you have a feature that contributes a `postStartCommand`. That command would be run in addition to whatever user-provided `postStartCommand` exists when the devcontainer starts up? 

This sounds like a great improvement to me. Features should be able to contribute partial devcontainer contents, so extending features to be able to contribute lifecycle hooks is a natural evolution of what features can do.

Chuxel:
[...]

Issue #2, opened 2022-01-21
Port features.json properties for most common runtime arguments


Chuxel:
As a part of development for dev container features (#60), we introduced properties for the most common runtime arguments used in `devcontainer.json`'s `runArgs` property. The great part about this implementation is that it worked with Docker Compose as well and is abstracted enough to work with other potential orchestrators or container runtimes down the road.  It would be awesome to bring these to `devcontainer.json` as well.

Specifically:

- `privileged`
- `capAdd`
- `securityOpt`
- `init`
- `mounts` in a revised form that is not quite so engine specific
- `entrypoint` (which would be useful with the proposed `overrideEntrypoint` from https://github.com/microsoft/vscode-remote-release/issues/3614#issuecomment-724125719)

We could also bring `containerEnv` into Docker Compose as it is in `features.json`.

Ideally we'd continue this trend whenever a new property along these lines is added to `features.json` (and vice versa).

Chuxel:
Yep! It should future proof the spec as well given interest in different container engines. These can have orchestrators that have subtle tweaks in how these need are set.  We don't want to "be the orchestrator", but provide a place to house subset of properties that are specific to dev that need to happen regardless of them - e.g. `capAdd` to enable ptrace based debuggers.

numeralz:
Is is possible to specify overrides for a feature added to _devcontainer.json_? It may solve a class of issues where the default arguments (like mounts) of feature definitions need to be removed/changed.

For example, I added the **docker-in-docker** feature to my _devcontainer.json_, but in order to be able to have docker-in-docker enabled in more than one containers on one host, you must mount `/var/lib/docker` to a uniquely-named volume. 
```json
	"features": {
		"docker-in-docker": {
			"version": "latest",
			"moby": true,
			"dockerDashComposeVersion": "v1"
			//FIXME: can we have "customizations": { "mounts": [  ] } or similar
		},
        },
        // attempt to mount /var/lib/docker to unique :
	"mounts": [
		"source=UNIQUE-dind-var-lib-docker,target=/var/lib/docker,type=volume"
	]
```
Fails because "docker-in-docker" feature def has a built-in mount to /var/lib/docker which conflicts with my uniquely-named volume:
from error log, formatted for readability:
```log
...
--mount source=UNIQUE-dind-var-lib-docker,target=/var/lib/docker,type=volume   <-- my attempt
--mount type=volume,src=dind-var-lib-docker,dst=/var/lib/docker    <-- built-in
[...]

PR #95, opened 2022-09-14
Image Metadata Proposal

chrmarti:
Continuing from https://github.com/devcontainers/spec/issues/18.

### DIFF
--- a/proposals/image-metadata.md
+++ b/proposals/image-metadata.md
@@ -4,7 +4,7 @@
 
 Record dev container config and feature metadata in prebuilt images, such that, the image and the built-in features can be used with a devcontainer.json (image-, Dockerfile- or Docker Compose-based) that does not repeat the dev container config or feature metadata. Other tools should be able to record the same metadata without necessarily using features themselves.
 
-Current dev container config that can be recorded in the image: ``
+Current dev container config that can be recorded in the image: `mounts`, `onCreateCommand`, `updateContentCommand`, `postCreateCommand`, `postStartCommand`, `postAttachCommand`, `customizations`, `remoteUser`, `userEnvProbe`, `remoteEnv`, `containerEnv`, `overrideCommand`, `portsAttributes`, `otherPortsAttributes`, `forwardPorts`, `shutdownAction`, `updateRemoteUserUID` and `hostRequirements`.
 
 Current feature metadata relevant for using the feature when it is already part of the image: `mounts`, `init`, `privileged`, `capAdd`, `securityOpt`, `entrypoint` and `customizations`.
 
--- a/proposals/image-metadata.md
+++ b/proposals/image-metadata.md
@@ -44,10 +44,10 @@ To apply the metadata together with a user's devcontainer.json at runtime the fo
 | Property | Type/Format | Merge Logic | devcontainer.json | Feature Metadata |
 | -------- | ----------- | ----------- | :---------------: | :--------------: |
 | `id` | E.g., `ghcr.io/devcontainers/features/node:1` | Not merged. |   | x |
-| `init` | `boolean` | `true` if at least one is `true`, `false` otherwise. |   | x |
-| `privileged` | `boolean` | `true` if at least one is `true`, `false` otherwise. |   | x |
-| `capAdd` | `string[]` | Union of all `capAdd` arrays without duplicates. |   | x |
-| `securityOpt` | `string[]` | Union of all `securityOpt` arrays without duplicates. |   | x |
+| `init` | `boolean` | `true` if at least one is `true`, `false` otherwise. | x | x |
+| `privileged` | `boolean` | `true` if at least one is `true`, `false` otherwise. | x | x |
+| `capAdd` | `string[]` | Union of all `capAdd` arrays without duplicates. | x | x |
+| `securityOpt` | `string[]` | Union of all `securityOpt` arrays without duplicates. | x | x |
 | `entrypoint` | `string` | Collected list of all entrypoints. |   | x |
 | `mounts` | `(string \| { type, src, dst })[]` | Collected list of all mountpoints. Conflicts: Last source wins. | x | x |
 | `onCreateCommand` | `string \| string[]` | Collected list of all onCreateCommands. | x |   |
@@ -71,6 +71,10 @@ To apply the metadata together with a user's devcontainer.json at runtime the fo
 | `hostRequirements` | `cpus`, `memory`, `storage` | Max value wins. | x |   |
 
 Variables in string values will be substituted at the time the value is applied. When the order matters, the devcontainer.json is considered last.
+
+## Additional devcontainer.json Properties
+
+We are adding support for `mounts`, `containerEnv`, `containerUser`, `init`, `privileged`, `capAdd`, and `securityOpt` to the devcontainer.json (also with Docker Compose) the same way these properties are already supported in the feature metadata.
 
 ## Notes
 
--- a/proposals/image-metadata.md
+++ b/proposals/image-metadata.md
@@ -55,7 +55,9 @@ To apply the metadata together with a user's devcontainer.json at runtime the fo
 | `postCreateCommand` | `string \| string[]` | Collected list of all postCreateCommands. | x |   |
 | `postStartCommand` | `string \| string[]` | Collected list of all postStartCommands. | x |   |
 | `postAttachCommand` | `string \| string[]` | Collected list of all postAttachCommands. | x |   |
+| `waitFor` | enum | Last value wins. | x |   |
 | `customizations` | Object of tool-specific customizations. | Merging is left to the tools. | x | x |
+| `containerUser` | `string` | Last value wins. | x |   |
 | `remoteUser` | `string` | Last value wins. | x |   |
 | `userEnvProbe` | `string` (enum) | Last value wins. | x |   |
 | `remoteEnv` | Object of strings. | Per variable, last value wins. | x |   |
@@ -68,7 +70,7 @@ To apply the metadata together with a user's devcontainer.json at runtime the fo
 | `updateRemoteUserUID` | `boolean` | Last value wins. | x |   |
 | `hostRequirements` | `cpus`, `memory`, `storage` | Max value wins. | x |   |
 
-Variables in string values will be substituted at the time the value is applied.
+Variables in string values will be substituted at the time the value is applied. When the order matters, the devcontainer.json is considered last.
 
 ## Notes
 
--- a/proposals/image-metadata.md
+++ b/proposals/image-metadata.md
@@ -1,0 +1,81 @@
+# Image Metadata
+
+## Goal
+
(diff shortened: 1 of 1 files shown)


proposals/image-metadata.md
```diff
@@ -0,0 +1,81 @@
+# Image Metadata
+
+## Goal
+
+Record dev container config and feature metadata in prebuilt images, such that, the image and the built-in features can be used with a devcontainer.json (image-, Dockerfile- or Docker Compose-based) that does not repeat the dev container config or feature metadata. Other tools should be able to record the same metadata without necessarily using features themselves.
+
+Current dev container config that can be recorded in the image: ``
+
+Current feature metadata relevant for using the feature when it is already part of the image: `mounts`, `init`, `privileged`, `capAdd`, `securityOpt`, `entrypoint` and `customizations`.
```
Chuxel, line 9: Is this trying to record current state?  I wasn't quite sure if this was trying to get at the relevant properties that would apply to image labels or just how things are now.

proposals/image-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+		}
+	},
+	...
+]
+```
+
+To simplify adding this metadata for other tools, we also support having a single top-level object with the same properties.
+
+The metadata is added to the image as a `devcontainer.metadata` label with a JSON string value representing the above array or single object.
+
```
Chuxel, line 39: I would also propose we bring the properties we'd support as a label in to devcontainer.json as well (as described in #2). 

We'd want to discourage the use of things like `runArgs` as this proposal comes together... otherwise we'd likely end up needing to include that too. It feels like that ends up being a part of this overall proposal because it.

Chuxel (commented):
Couple thoughts/questions

proposals/image-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+| `portsAttributes` | Map of ports to attributes. | Per port (not per port attribute), last value wins. | x |   |
+| `otherPortsAttributes` | Port attributes. | Last value wins (not per port attribute). | x |   |
+| `forwardPorts` | `(number \| string)[]` | Union of all ports without duplicates. Last one wins (when mapping changes). | x |   |
+| `shutdownAction` | `string` (enum) | Last value wins. | x |   |
+| `updateRemoteUserUID` | `boolean` | Last value wins. | x |   |
+| `hostRequirements` | `cpus`, `memory`, `storage` | Max value wins. | x |   |
+
+Variables in string values will be substituted at the time the value is applied.
+
+## Notes
```
joshspicer, line 73: How do we communicate to the user (someone using Codespaces or Remote-Containers) that the image they are referencing in the `image` property is "special" and will be applying additional dev container configuration into their dev environment?

After this change is applied, images under `mcr.microsoft.com/devcontainers` will behave differently than before (eg: the `remoteUser` will be applied in cases where it previously didn't and the user was defaulted to root). 

Just something to think about 🤔

proposals/image-metadata.md
chrmarti: Current state. Apparently forgot to paste in the list for the devcontainer.json above. (We could remove these as we repeat them in the table at the end.)

proposals/image-metadata.md
chrmarti: I would prefer to keep this change scoped to the metadata on the image. We can always add more properties later.

proposals/image-metadata.md
chrmarti: I agree. We could show additional information in a hover on the "dockerfile" or "image" properties in the devcontainer.json.

proposals/image-metadata.md
chrmarti: Actually: "Current state" in the sense that these are the currently existing properties we will record in the new image metadata. :)

Chuxel: Yeah that was kind of what I was thinking too - the table at the end covers it.

proposals/image-metadata.md
joshspicer: There's this really slick looking, non-intrusive "info" diagnostic that vscode provides (the three grey dots underneath the string)

image

Would be cool to provide info here on the config that will be applied!

joshspicer (approved):

proposals/image-metadata.md
joshspicer: this and some useful logging about applying the labels as config might be enough

jkeech (approved):

proposals/image-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+| `onCreateCommand` | `string \| string[]` | Collected list of all onCreateCommands. | x |   |
+| `updateContentCommand` | `string \| string[]` | Collected list of all updateContentCommands. | x |   |
+| `postCreateCommand` | `string \| string[]` | Collected list of all postCreateCommands. | x |   |
+| `postStartCommand` | `string \| string[]` | Collected list of all postStartCommands. | x |   |
+| `postAttachCommand` | `string \| string[]` | Collected list of all postAttachCommands. | x |   |
+| `customizations` | Object of tool-specific customizations. | Merging is left to the tools. | x | x |
+| `remoteUser` | `string` | Last value wins. | x |   |
+| `userEnvProbe` | `string` (enum) | Last value wins. | x |   |
+| `remoteEnv` | Object of strings. | Per variable, last value wins. | x |   |
+| `containerEnv` | Object of strings. | Per variable, last value wins. | x |   |
```
samruddhikhandale, line 62: Isn't `containerEnv` supported in `Feature Metadata`? We do have it in `devcontainer-feature.json` in almost all of the official Features.

samruddhikhandale: Provided same `containerEnv` variable is present in `devcontainer.json` and feature metadata, should the value in `devcontainer.json` win ?

proposals/image-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+
+| Property | Type/Format | Merge Logic | devcontainer.json | Feature Metadata |
+| -------- | ----------- | ----------- | :---------------: | :--------------: |
+| `id` | E.g., `ghcr.io/devcontainers/features/node:1` | Not merged. |   | x |
+| `init` | `boolean` | `true` if at least one is `true`, `false` otherwise. |   | x |
+| `privileged` | `boolean` | `true` if at least one is `true`, `false` otherwise. |   | x |
+| `capAdd` | `string[]` | Union of all `capAdd` arrays without duplicates. |   | x |
+| `securityOpt` | `string[]` | Union of all `securityOpt` arrays without duplicates. |   | x |
+| `entrypoint` | `string` | Collected list of all entrypoints. |   | x |
+| `mounts` | `(string \| { type, src, dst })[]` | Collected list of all mountpoints. Conflicts: Last source wins. | x | x |
```
samruddhikhandale, line 52: As `mounts` exist in `devcontainer.json` and `feature metadata`, and given that we have conflict in them, which would be last source in that case? should it be `devcontainer.json` as that's something the image author has control over rather than Features which could written by someone else?

proposals/image-metadata.md
chrmarti: Currently `containerEnv` from the feature metadata is added to the Dockerfile while `containerEnv` from the devcontainer.json is applied when the container starts. To also support variables like `${localEnv:VAR_NAME}` in the feature metadata's `containerEnv` it would make sense to apply these when the containers starts.

@edgonmsft @joshspicer We at one point moved the `ENV` statements in the Dockerfile before running the feature install scripts. Would moving these `ENV` statements out cause issues that cannot be addressed in a different way?

proposals/image-metadata.md
chrmarti: I agree. Will add a note that when order matters, the devcontainer.json is considered last.

proposals/image-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+| `onCreateCommand` | `string \| string[]` | Collected list of all onCreateCommands. | x |   |
+| `updateContentCommand` | `string \| string[]` | Collected list of all updateContentCommands. | x |   |
+| `postCreateCommand` | `string \| string[]` | Collected list of all postCreateCommands. | x |   |
+| `postStartCommand` | `string \| string[]` | Collected list of all postStartCommands. | x |   |
+| `postAttachCommand` | `string \| string[]` | Collected list of all postAttachCommands. | x |   |
+| `customizations` | Object of tool-specific customizations. | Merging is left to the tools. | x | x |
+| `remoteUser` | `string` | Last value wins. | x |   |
```
chrmarti, line 59: I think we missed `containerUser` and `waitFor`. Will add these.

proposals/image-metadata.md
chrmarti: We agreed to keep it as-is. 👍

samruddhikhandale (approved):
LGTM ✨

proposals/image-metadata.md
chrmarti: Adding below.

proposals/image-metadata.md
```diff
@@ -0,0 +1,87 @@
...
+| `forwardPorts` | `(number \| string)[]` | Union of all ports without duplicates. Last one wins (when mapping changes). | x |   |
+| `shutdownAction` | `string` (enum) | Last value wins. | x |   |
+| `updateRemoteUserUID` | `boolean` | Last value wins. | x |   |
+| `hostRequirements` | `cpus`, `memory`, `storage` | Max value wins. | x |   |
+
+Variables in string values will be substituted at the time the value is applied. When the order matters, the devcontainer.json is considered last.
+
+## Additional devcontainer.json Properties
+
+We are adding support for `mounts`, `containerEnv`, `containerUser`, `init`, `privileged`, `capAdd`, and `securityOpt` to the devcontainer.json (also with Docker Compose) the same way these properties are already supported in the feature metadata.
```
chrmarti, line 77: @Chuxel Not adding `entrypoint` to the devcontainer.json because I think that should not be "additive" like `entrypoint` from features. I suggest we discuss and spec the details of this in https://github.com/devcontainers/spec/issues/2.

Chuxel: Yeah, the proposal for adding other "commands" into features (#60) would make them all additive, so I think this would be consistent with that - it just runs as the container user while other things run as the remote user.

chrmarti: Not sure I understand: Are you proposing to make `entrypoint` in the devcontainer.json add to any existing entrypoint/entrypoints? I was thinking that `entrypoint` in the devcontainer.json should override the image's entrypoint to get the same behavior as users are used to from the Dockerfile and docker-compose.yml. Making it additive would come as a surprise for users.

Chuxel: Hmmm.  I guess that is the question, you are right. Typically when I have been using the entrypoint in a Dockerfile, it hasn't been to override it as much as to start something up as the container user on "boot".

Ok, yeah that makes sense to discuss in #2 more.

Chuxel (approved):

Issue #62, opened 2022-07-15
Support for adding unique ID into volume mount names


Chuxel:
This proposal is an enhancement idea to the current [dev container features proposal](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md).

---
This spec proposal is a port of the proposed solution to https://github.com/microsoft/vscode-remote-release/issues/5679 and relates to https://github.com/devcontainers/spec/issues/2#issuecomment-1185755881.

For both Dev Container Features (#61) and normal devcontainer.json scenarios, it is sometimes necessary to have a unique identifier generated that can be referenced in different places.  

The most critical example of this is named volumes. When you have multiple dev containers running in the same environment (e.g. locally), the docker-in-docker feature needs to be able to create unique volume names for each to store docker data. Any feature that references a mount point will face the same challenge.

The proposed solution is to introduce a variable like `${devcontainerId}` that is a unique identifier for the resulting container. The format of the identifier does not matter as much as the fact it can be broadly used. So generally it should be alpha numeric. Features can then reference this identifier where appropriate (like in mounts).

@alexdima @chrmarti @jkeech @edgonmsft - This has been the most consistent ask from early adopters of dev container features to date.  So I think we'll want to cover it in the spec.

stuartleeks:
Is there any consensus on how the `devcontainerId` value would be generated?

My initial thought when thinking about this a while back was to use the `name` from `devcontainer.json`. The downside is that changing the `name` would break the volume links, which isn't ideal. Especially as I often start working with a dev container for a while before noticing that I haven't put something meaningful in the `name` value so end up changing it later to something to help identify the project.

Another idea is to add an `id` property to `devcontainer.json` and have the CLI/extension populate it with a new randomly generated value if it doesn't exist. This could then be the value used for `devcontainerId` and users could still rename the dev container without breaking the volume links in features.

stuartleeks:
An alternative to generating the `id` and modifying a user's `devcontainer.json` would be to use `id` if it exists and `name` (or a derivative of it) as a fallback. In this case, it might help steer into the pit of success if the extension auto-generated a random `id` value when a user adds a `devcontainer.json` via the "Remote-Containers: Add Development Container Configuration Files..." command?

chrmarti:
I would use a similar approach to how we name images which uses `vsc-${workspaceFolderBasename}-${workspaceFolderHash}`.

[...]

PR #96, opened 2022-09-14
Dev Container Id Proposal

chrmarti:
Continuing from https://github.com/devcontainers/spec/issues/62.


Chuxel (approved):

joshspicer (approved):

proposals/devcontainer-id-variable.md
```diff
@@ -0,0 +1,60 @@
+## Goal
+
+Allow features to refer to an identifier that is unique to the dev container they are installed into and that is stable across rebuilds.
+
+E.g., the `docker-in-docker` feature needs a way to mount a volume per dev container.
+
+## Proposal
+
+The identifier will be referred to as `${devcontainerId}` in the feature metadata and that will be replaced with the dev container's id. It should only be used in parts of the metadata that is not used for building the image because that would otherwise prevent pre-building the image at a time when the dev container's id is not known yet. Excluding boolean and enum properties the properties supporting `${devcontainerId}` are: `entrypoint`, `mounts`, `customizations`.
```
jkeech, line 9: Does it make sense to allow `devcontainerId` to be passed into lifecycle hooks in some way? I'm not aware of a need for this currently, so we could always defer it and add it later once there is a know use case.

proposals/devcontainer-id-variable.md
```diff
@@ -0,0 +1,60 @@
...
+
+Implementations can choose how to compute this identifier. They must ensure that it is unique among other dev containers on the same Docker host and that it is stable across rebuilds of dev containers. The identifier must only contain alphanumeric characters. We describe a way to do this below.
```
jkeech, line 11: 👍

proposals/devcontainer-id-variable.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Label-based Implementation 
+
+The following assumes that a dev container can be identified among other dev containers on the same Docker host by a set of labels on the container. Implementations may choose to follow this approach.
+
+The identifier is derived from the set of container labels uniquely identifying the dev container. It is up to the implementation to choose these labels, current experience suggests that the labels mainly refer to the dev container config. E.g., if the dev container is based on a local folder the label could be named `devcontainer.local_folder` and have the local folder's path as its value.
```
jkeech, line 17: Recommend to remove this part since it doesn't seem to make it clearer to me.
```suggestion
The identifier is derived from the set of container labels uniquely identifying the dev container. It is up to the implementation to choose these labels. E.g., if the dev container is based on a local folder the label could be named `devcontainer.local_folder` and have the local folder's path as its value.
```

samruddhikhandale (approved):
LGTM :sparkles

edgonmsft (approved):
LGTM

proposals/devcontainer-id-variable.md
```diff
@@ -0,0 +1,60 @@
...
+
+E.g., the `docker-in-docker` feature could use the dev container id with:
```
jkeech, line 19: To avoid confusion with the syntax for feature ids (publishing syntax vs consuming syntax), it might be nice to point to the published docker-in-docker feature for this example instead of just referring to the name `docker-in-docker`.

```suggestion
E.g., the [`ghcr.io/devcontainers/features/docker-in-docker` feature](https://github.com/devcontainers/features/blob/main/src/docker-in-docker/devcontainer-feature.json) could use the dev container id with:
```

jkeech (approved):
Looks good overall. I left a couple small suggestions/questions.

proposals/devcontainer-id-variable.md
chrmarti: That makes sense. (As discussed before: The only limitation is that it is not available when building the image because that can be prebuilt in CI.)

Issue #22, opened 2022-03-25
Addition of "extends" top level property to enable simple configuration inheritance


greggroth:
# Problem

Multiple teams collaborating on a common codebase may have different dependency or setup needs and currently this is done by sharing a single `devcontainer.json` configuration with all of their individual needs combined. https://github.com/microsoft/dev-container-spec/issues/6 describes support for multiple configuration files, but there isn't set a way to consolidate the shared configuration into a single file.

# Proposed Solution

Introduce a new top level key `"extends"` with a value that is a relative a file path within the same repository to a "parent" devcontainer configuration. The configurations will be merged using the [same rules applied by a Docker Compose overrides file](https://docs.docker.com/compose/extends/#adding-and-overriding-configuration).

```json5
// .devcontainer/defaults.json
{
    "name": "example/project",
    "forwardPorts": [80, 5432],
    "hostRequirements": {
        "storage": "64gb",
        "memory": "16gb"
    }
}

// .devcontainer/project.json
{
    "extends": "defaults.json",
    "forwardPorts": [2222],
    "hostRequirements": {
        "memory": "32gb"
    },
   "onCreateCommand": ".devcontainer/on-create-command.sh",
}

// Results in
{
    "name": "example/project",
    "forwardPorts": [80, 5432, 2222],    // <-- Array values are the UNION
    "hostRequirements": {
        "storage": "64gb",
        "memory": "32gb"                          // <-- Basic types overwrite
    },
   "onCreateCommand": ".devcontainer/on-create-command.sh",   // <-- New keys are added
}
```

# Future 

## Add support for referring to configuration outside the repository

Users may have a use-case for keeping some shared configuration in a separate repository. The described solution does not support this, but support for this could be added.

## Add support for JSON Schema/Open API 3.0 style "$ref" document imports

The described solution is opinionated about how to merge a document, which may not be a good fit for everyone's needs (see https://github.com/docker/compose/issues/3729). A more sophisticated document reference method could be introduced to give users more control to import documents within objects or arrays and to reference objects within imported documents. For example: 

Given a file "defaults.json":

```json5
// .devcontainer/defaults.json
{
    "name": "microsoft/foo",
    "extensions": [
        "/root/hello-githug.vsix"
    ],
    "forwardPorts": [80, 5432]
    "hostRequirements": {
        "storage": "64gb",
        "memory": "32gb"
    },
    "portsAttributes": {
        "80": {
            "label": "web"
        },
        "5432": {
[...]

PR #98, opened 2022-09-15
Dev container Templates & distribution

samruddhikhandale:
The proposal is an idea of how to author and distribute Templates. This follows a very similar mechanism as of dev container [Features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md)

### DIFF
--- a/proposals/devcontainer-templates-distribution.md
+++ b/proposals/devcontainer-templates-distribution.md
@@ -23,9 +23,7 @@ Source code for the set follows the example file structure below:
 ├── src
 │   ├── dotnet
 │   │   ├── devcontainer-template.json
-│   │   ├── .devcontainer
-│   │       ├── devcontainer.json
-│   │       └── ...
+│   │   ├── .devcontainer.json
 │   │   ├── ...
 |   ├
 │   ├── docker-from-docker
@@ -46,19 +44,17 @@ Source code for the set follows the example file structure below:
 │   │   ├── ...
 ```
 
-...where `src` is a directory containing a sub-folder with the name of the Template (e.g. `src/dotnet` or `src/docker-from-docker`) with at least a file named `devcontainer-template.json` that contains the Template metadata, and a `.devcontainer` that the supporting tools will drop into an existing project or folder.
+...where `src` is a directory containing a sub-folder with the name of the Template (e.g. `src/dotnet` or `src/docker-from-docker`) with at least a file named `devcontainer-template.json` that contains the Template metadata, and a `.devcontainer.json` (or `.devcontainer/devcontainer.json`) that the supporting tools will drop into an existing project or folder.
 
 Each sub-directory should be named such that it matches the `id` field of the `devcontainer-template.json`.  Other files can also be included in the Templates's sub-directory, and will be included during the [packaging step](#packaging) alongside the two required files.  Any files that are not part of the Templates's sub-directory (e.g. outside of `src/dotnet`) will not included in the [packaging step](#packaging).
 
 ## Versioning
 
-Each Template is individually [versioned according to the semver specification](https://semver.org/).  The `version` property in the respective `devcontainer-template.json` file is parsed to determine if the Template should be republished.
-
-Tooling that handles publishing Templates will not republish Templates if that exact version has already been published; however, tooling must republish major and minor versions in accordance with the semver specification.
+Each Template is published with only the `latest` tag. Tooling that handles releasing Templates will republish the `latest` tag every time a new release is created.
 
 ## Packaging
 
-Templates are distributed as tarballs. The tarball contains the entire contents of the Template sub-directory, including the `devcontainer-template.json`, `.devcontainer`, and any other files in the directory.
+Templates are distributed as tarballs. The tarball contains the entire contents of the Template sub-directory, including the `devcontainer-template.json`, `.devcontainer.json` (or `.devcontainer/devcontainer.json`), and any other files in the directory.
 
 The tarball is named `devcontainer-template-<id>.tgz`, where `<id>` is the Templates's `id` field.
 
@@ -85,30 +81,27 @@ A user can add the Templates in to their projects as defined by the [Templates S
 
 An OCI registry that implements the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec) serves as the primary distribution mechanism for Templates.
 
-Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:version]`, where version is the major, minor, and patch version of the feature, according to the semver specification.
+Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:latest]`.
 
 > **Note:** The `namespace` is a unique indentifier for the collection of templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
 
 A custom media type `application/vnd.devcontainers` and `application/vnd.devcontainers.layer.v1+tar` are used as demonstrated below.
 
-For example, the `go` Template in the `devcontainers/templates` namespace at version `1.2.3` would be pushed to the ghcr.io OCI registry.  
+For example, the `go` Template in the `devcontainers/templates` namespace would be pushed to the ghcr.io OCI registry with the `latest` tag. 
 
 > **Note:** The example below uses [`oras`](https://oras.land/) for demonstration purposes.  A supporting tool should directly implement the required functionality from the aforementioned OCI artifact distribution specification.
 
 ```bash
-# ghcr.io/devcontainers/templates/go:1 
+# ghcr.io/devcontainers/templates/go:latest
 REGISTRY=ghcr.io
 NAMESPACE=devcontainers/templates
 TEMPLATE=go
 
 ARTIFACT_PATH=devcontainer-template-go.tgz
 
-for VERSION in 1  1.2  1.2.3  latest
-do
-    oras push ${REGISTRY}/${NAMESPACE}/${TEMPLATE}:${VERSION} \
-            --manifest-config /dev/null:application/vnd.devcontainers \
-                             ./${ARTIFACT_PATH}:application/vnd.devcontainers.layer.v1+tar
-done
+oras push ${REGISTRY}/${NAMESPACE}/${TEMPLATE}:latest \
+        --manifest-config /dev/null:application/vnd.devcontainers \
+                ./${ARTIFACT_PATH}:application/vnd.devcontainers.layer.v1+tar
 
 ```
 
--- a/proposals/devcontainer-templates-distribution.md
+++ b/proposals/devcontainer-templates-distribution.md
@@ -9,13 +9,13 @@ Goals include:
 
 ## Source code
 
-Templates source code is stored in a git repository.
+A Template's source code is stored in a git repository.
 
 For ease of authorship and maintenance, [1..n] Templates can share a single git repository. This set of Templates is referred to as a "collection," and will share the same [`devcontainer-collection.json`](#devcontainer-collection.json) file and "namespace" (eg. `<owner>/<repo>`).
 
 > **Note:** Templates and [Features](./devcontainer-features.md) should be placed in different git repositories. 
 
-Source code for the set follows the example file structure below:
+Source code for a set of Templates follows the example file structure below:
 
 ```
 .
@@ -69,13 +69,13 @@ The `devcontainer-collection.json` is an auto-generated metadata file.
 | `sourceInformation` | object | Metadata from the implementing packaging tool. |
 | `templates` | array | The list of Templates that are contained in this collection.|
 
-Each templates's `devcontainer-template.json` metadata file is appended into the `templates` top-level array.
+Each Template's `devcontainer-template.json` metadata file is appended into the `templates` top-level array.
 
 ## Distribution
 
-There are several supported ways to distribute templates.  Distribution is handled by the implementing packaging tool.
+There are several supported ways to distribute Templates.  Distribution is handled by the implementing packaging tool.
 
-A user can add the Templates in to their projects as defined by the [Templates Supporting Tools](./devcontainer-templates.md/#adding-a-template-to-a-project-or-codespace).
+A user can add a Template in to their projects as defined by the [supporting Tools](../docs/specs/supporting-tools.md#supporting-tools-and-services).
 
 ### OCI Registry
 
@@ -83,7 +83,7 @@ An OCI registry that implements the [OCI Artifact Distribution Specification](ht
 
 Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:latest]`.
 
-> **Note:** The `namespace` is a unique identifier for the collection of templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
+> **Note:** The `namespace` is a unique identifier for the collection of Templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
 
 A custom media type `application/vnd.devcontainers` and `application/vnd.devcontainers.layer.v1+tar` are used as demonstrated below.
 
@@ -105,7 +105,7 @@ oras push ${REGISTRY}/${NAMESPACE}/${TEMPLATE}:latest \
 
 ```
 
-The "namespace" is the globally identifiable name for the collection of templates. (eg: `owner/repo` for the source code's git repository).
+The "namespace" is the globally identifiable name for the collection of Templates. (eg: `owner/repo` for the source code's git repository).
 
 The auto-generated `devcontainer-collection.json` is pushed to the registry with the same `namespace` as above and no accompanying `template` name. The collection file is always tagged as `latest`.
 
--- a/proposals/devcontainer-templates-distribution.md
+++ b/proposals/devcontainer-templates-distribution.md
@@ -50,7 +50,9 @@ Each sub-directory should be named such that it matches the `id` field of the `d
 
 ## Versioning
 
-Each Template is published with only the `latest` tag. Tooling that handles releasing Templates will republish the `latest` tag every time a new release is created.
+Each Template is individually [versioned according to the semver specification](https://semver.org/). The `version` property in the respective `devcontainer-template.json` file is parsed to determine if the Template should be republished.
+
+Tooling that handles publishing Templates will not republish Templates if that exact version has already been published; however, tooling must republish major and minor versions in accordance with the semver specification.
 
 ## Packaging
 
@@ -81,27 +83,30 @@ A user can add a Template in to their projects as defined by the [supporting Too
 
 An OCI registry that implements the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec) serves as the primary distribution mechanism for Templates.
 
-Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:latest]`.
+Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:version]`, where version is the major, minor, and patch version of the Template, according to the semver specification.
 
 > **Note:** The `namespace` is a unique identifier for the collection of Templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
 
 A custom media type `application/vnd.devcontainers` and `application/vnd.devcontainers.layer.v1+tar` are used as demonstrated below.
 
-For example, the `go` Template in the `devcontainers/templates` namespace would be pushed to the ghcr.io OCI registry with the `latest` tag. 
+For example, the `go` Template in the `devcontainers/templates` namespace at version `1.2.3` would be pushed to the ghcr.io OCI registry.
 
 > **Note:** The example below uses [`oras`](https://oras.land/) for demonstration purposes.  A supporting tool should directly implement the required functionality from the aforementioned OCI artifact distribution specification.
 
 ```bash
-# ghcr.io/devcontainers/templates/go:latest
+# ghcr.io/devcontainers/templates/go:1
 REGISTRY=ghcr.io
 NAMESPACE=devcontainers/templates
 TEMPLATE=go
 
 ARTIFACT_PATH=devcontainer-template-go.tgz
 
-oras push ${REGISTRY}/${NAMESPACE}/${TEMPLATE}:latest \
-        --manifest-config /dev/null:application/vnd.devcontainers \
-                ./${ARTIFACT_PATH}:application/vnd.devcontainers.layer.v1+tar
+for VERSION in 1  1.2  1.2.3  latest
+do
+        oras push ${REGISTRY}/${NAMESPACE}/${TEMPLATE}:${VERSION} \
+                --manifest-config /dev/null:application/vnd.devcontainers \
+                        ./${ARTIFACT_PATH}:application/vnd.devcontainers.layer.v1+tar
+done
 
 ```
 
--- a/proposals/devcontainer-templates-distribution.md
+++ b/proposals/devcontainer-templates-distribution.md
@@ -1,0 +1,127 @@
+# Distribution and Discovery of Dev Container Templates
+
+This specification defines a pattern where community members and organizations can author and self-publish [Dev Container Templates](./devcontainer-templates.md). 
+
+Goals include:
+
+- For Template authors, create a "self-service" way to publish a Template, either publicly or privately, that is not centrally controlled.
+- Provide the ability to standardize publishing such that supporting tools may implement their own mechanism to aid Template discoverability as they see fit.
+
+## Source code
+
+Templates source code is stored in a git repository.
+
+For ease of authorship and maintenance, [1..n] Templates can share a single git repository. This set of Templates is referred to as a "collection," and will share the same [`devcontainer-collection.json`](#devcontainer-collection.json) file and "namespace" (eg. `<owner>/<repo>`).
+
+> **Note:** Templates and [Features](./devcontainer-features.md) should be placed in different git repositories. 
+
+Source code for the set follows the example file structure below:
+
+```
+.
+├── README.md
+├── src
+│   ├── dotnet
+│   │   ├── devcontainer-template.json
+│   │   ├── .devcontainer
+│   │       ├── devcontainer.json
+│   │       └── ...
+│   │   ├── ...
+|   ├
+│   ├── docker-from-docker
+│   │   ├── devcontainer-template.json
+│   │   ├── .devcontainer
+│   │       ├── devcontainer.json
+│   │       ├── Dockerfile
+│   │       └── ...
+│   │   ├── ...
+|   ├
+│   ├── go-postgres
+│   │   ├── devcontainer-template.json
+│   │   ├── .devcontainer
+│   │       ├── devcontainer.json
+│   │       ├── docker-compose.yml
+│   │       ├── Dockerfile
+│   │       └── ...
+│   │   ├── ...
+```
+
+...where `src` is a directory containing a sub-folder with the name of the Template (e.g. `src/dotnet` or `src/docker-from-docker`) with at least a file named `devcontainer-template.json` that contains the Template metadata, and a `.devcontainer` that the supporting tools will drop into an existing project or folder.
+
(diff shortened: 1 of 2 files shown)


proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,127 @@
...
+│   │   ├── devcontainer-template.json
+│   │   ├── .devcontainer
+│   │       ├── devcontainer.json
+│   │       ├── docker-compose.yml
+│   │       ├── Dockerfile
+│   │       └── ...
+│   │   ├── ...
+```
+
+...where `src` is a directory containing a sub-folder with the name of the Template (e.g. `src/dotnet` or `src/docker-from-docker`) with at least a file named `devcontainer-template.json` that contains the Template metadata, and a `.devcontainer` that the supporting tools will drop into an existing project or folder.
```
chrmarti, line 49: Do we also want to support the `.devcontainer.json` file in the workspace root folder?

chrmarti: Also: Probably need to change `.devcontainer` to `.devcontainer/devcontainer.json` in the text to fully make sense.

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,144 @@
...
+      "default": "value1"
+    }
+  }
+}
+```
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| `optionId` | string | ID of the option that is converted into an all-caps and snake case build argument with the selected value in it. |
+| `optionId.type` | string | Type of the option. Valid types are currently: `boolean`, `string` |
```
chrmarti, line 63: Build args are always strings (similar to env variables).

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,144 @@
...
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| `id` | string | ID of the Template. The `id` should be unique in the context of the repository/published package where the Template exists and must match the name of the directory where the `devcontainer-template.json` resides. |
+| `version` | string | The semantic version of the Template. |
+| `name` | string | Name of the Template. |
+| `description` | string | Description of the Template. |
+| `documentationURL` | string | Url that points to the documentation of the Template. |
+| `licenseURL` | string | Url that points to the license of the Template. |
+| `type` | string | Type of the dev container (image, dockerfile, dockerCompose) created by the Template. |
```
chrmarti, line 36: How is this used?

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,144 @@
...
+| `image` | string | Name of an image in a container registry ([DockerHub](https://hub.docker.com/), [GitHub Container Registry](https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages), [Azure Container Registry](https://azure.microsoft.com/en-us/products/container-registry/)) that the [Template supporting services / tools](#template-supporting-tools-and-services) should use to populate list of image variants. |
```
chrmarti, line 37: Is this for the UI? Can `options` be used for this instead of the additional `image` property?

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,144 @@
...
+| `features` | object | List of [Feature id](devcontainer-features.md/#referencing-a-feature) that the [Template supporting services / tools](#template-supporting-tools-and-services) should use to populate different configuration options. |
```
chrmarti, line 38: Is this for showing a list to the user to pick from? Would it include the major version?

chrmarti (changes requested):
Looks great overall. Left a few questions.

proposals/devcontainer-templates-distribution.md
samruddhikhandale: > Do we also want to support the .devcontainer.json file in the workspace root folder?
Yep, missed that. Added

proposals/devcontainer-templates.md
samruddhikhandale: Currently, the `definitions` have this `type` property (`singleContainer` and `dockerCompose`) which is used by the UI to determine how to start or stop the container and configure few other things. Hence, added it here. 

Also, as per today's discussion, this will also be useful for listing Templates and people can sort with the `type`

proposals/devcontainer-templates.md
samruddhikhandale: Removed as per the discussion and added to `options` instead.

proposals/devcontainer-templates.md
samruddhikhandale: Removed as per the discussion and added to options instead.

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,169 @@
...
+A Template is a self contained entity in a folder with at least a `devcontainer-template.json` and [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson).  Additional files are permitted and are packaged along side the required files.
+
+```
++-- template
+|    +-- devcontainer-template.json
+|    +-- .devcontainer.json
+|    +-- (other files)
+```
+
+## devcontainer-template.json properties
```
samruddhikhandale, line 18: @Chuxel  For the `publisher` info which is needed by the UI -  I believe we will have a similar index for Templates as we have for [Features](https://github.com/devcontainers/devcontainers.github.io/blob/gh-pages/_data/collection-index.yml#L2) and this index will be used by the tooling to display list of Templates. We can then make use of `maintainer` property to display to `publisher/maintainer` name wherever needed.  

 Adding `publisher` prop to `devcontainer-template.json` feels a little unsafe as any author can claim to be someone else and we currently don't have identity checking mechanism in place. At least, with the search index, we will manually look at the PR created by the author (to add his Templates to crawler) and confirm if the maintainer prop matches with  other details.

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+A user can add the Templates in to their projects as defined by the [Templates Supporting Tools](./devcontainer-templates.md/#adding-a-template-to-a-project-or-codespace).
+
+### OCI Registry
+
+An OCI registry that implements the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec) serves as the primary distribution mechanism for Templates.
+
+Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:latest]`.
+
+> **Note:** The `namespace` is a unique indentifier for the collection of templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`.
```
joshspicer, line 86:
```suggestion
> **Note:** The `namespace` is a unique identifier for the collection of templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,178 @@
+# Dev Container Templates reference
+
+Development container "Templates" are self-contained, shareable units of development container configuration. The name comes from the idea that adding a Template to your project helps it get up and running with a containerized environment. The Templates describe the appropriate container image, [Features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md), runtime arguments for starting the container, and VS Code extensions that should be installed. Each provides a container configuration file ([`.devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson)) and other needed files that you can drop into any existing folder as a starting point for containerizing your project.
```
joshspicer, line 3:
```suggestion
Development container "Templates" are source files packaged together that encode configuration for a complete development environment. A Template can be used in a new or existing project.  A [supporting tool](https://containers.dev/supporting) will use the configuration from the template to build a development container.

Configuration is placed in a [`.devcontainer/devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson). The `devcontainer.json` can reference other files within the template, such as a `Dockerfile` or `docker-compose.yml`. 

A template can also provide additional example source files (eg: boilerplate code or a [lifecycle script](https://containers.dev/implementors/json_reference/#lifecycle-scripts).

```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,178 @@
...
+
+Template metadata is captured by a `devcontainer-template.json` file in the root folder of the Template.
+
+## Folder structure 
+
+A Template is a self contained entity in a folder with at least a `devcontainer-template.json` and [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson).  Additional files are permitted and are packaged along side the required files.
```
joshspicer, line 9:
```suggestion
A single Template is a folder with at least a `devcontainer-template.json` and [`devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson).  Additional files are permitted and are packaged along side the required files.
```

proposals/devcontainer-templates.md
joshspicer: Perhaps the tooling could add some of these attributes.  I imagine from some `devcontainer template package` command, it would not be too difficult to programatically add several of these attributes (and ensures consistency/doesn't require action from the author).

proposals/devcontainer-templates.md
samruddhikhandale: Thanks for the edit. looks good, added with little tweaks

proposals/devcontainer-templates.md
samruddhikhandale: Just wondering what if the tooling mistags some times ?

Example - lets say a Template contains all of three files (.devcontainer.json, dockerfile and dockercompose), the Template author writes code in such a way that the dev container solely depends on an image and has only added boiler plate code for `dockerfile` and `dockerCompose` or samples of how to configure using them. If the tooling simply determines the value based on the file presence of some other attributes, I wonder it might mis configure sometimes 🤔

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+### OCI Registry
+
+An OCI registry that implements the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec) serves as the primary distribution mechanism for Templates.
+
+Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:latest]`.
+
+> **Note:** The `namespace` is a unique identifier for the collection of templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
+
+A custom media type `application/vnd.devcontainers` and `application/vnd.devcontainers.layer.v1+tar` are used as demonstrated below.
```
chrmarti, line 88: @joshspicer Would it make sense to use a different media type for templates and features to distinguish the two? Does OCI's artifact spec have a recommendation?

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,192 @@
+# Dev Container Templates reference
+
+Development container "Templates" are source files packaged together that encode configuration for a complete development environment. A Template can be used in a new or existing project and a [supporting tool](https://containers.dev/supporting) will use the configuration from the template to build a development container.
+
+Configuration is placed in a [`.devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson). Alternatively, `.devcontainer/devcontainer.json` can also be used if the container needs to reference other files within the template, such as a `Dockerfile` or `docker-compose.yml`. A template can also provide additional source files (eg: boilerplate code or a [lifecycle script](https://containers.dev/implementors/json_reference/#lifecycle-scripts).
```
chrmarti, line 5: Can a `.devcontainer.json` also reference other files? (I thought so, but the wording seems to suggest otherwise.)

Prepend "The" to "configuration"?

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,192 @@
...
+
+and it has the following `.devcontainer.json` file:
+
+```json
+{
+	"name": "Java",
+	"image": "mcr.microsoft.com/devcontainers/java:0-{{imageVariant}}",
+	"features": {
+		"ghcr.io/devcontainers/features/node:1": {
+			"version": "{{nodeVersion}}",
```
chrmarti, line 156: We are using `${localEnv:<env var name>}` and similar variables at runtime. Should we use `${templateOption:<option id>}` (e.g., `${templateOption:nodeVersion}` here) for consistency?

chrmarti (changes requested):
Great progress. Left a few more questions.

proposals/devcontainer-templates-distribution.md
samruddhikhandale: Thinking of having a `label` to distinguish them, if that makes sense. 

> "labels": {
>     "com.github.package.type": "devcontainer_template"
>   }

For Features we already have it as `"com.github.package.type": "devcontainer_feature"`

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+│   │   ├── ...
+```
+
+...where `src` is a directory containing a sub-folder with the name of the Template (e.g. `src/dotnet` or `src/docker-from-docker`) with at least a file named `devcontainer-template.json` that contains the Template metadata, and a `.devcontainer.json` (or `.devcontainer/devcontainer.json`) that the supporting tools will drop into an existing project or folder.
+
+Each sub-directory should be named such that it matches the `id` field of the `devcontainer-template.json`.  Other files can also be included in the Templates's sub-directory, and will be included during the [packaging step](#packaging) alongside the two required files.  Any files that are not part of the Templates's sub-directory (e.g. outside of `src/dotnet`) will not included in the [packaging step](#packaging).
+
+## Versioning
+
+Each Template is published with only the `latest` tag. Tooling that handles releasing Templates will republish the `latest` tag every time a new release is created.
```
Chuxel, line 53: Hmmm. Not sure we landed on only publishing latest.  I think what we figured out is that we'd want versioning, but that we also need to factor in the schema version. This would allow us to do automated verion selection based on the schema much like VS Code extensions do today down the road. But you'd need to have some concept of versioning to make that work. Thoughts @chrmarti @alexdima?

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Template Supporting Tools and Services
+
+Outlines tools and services that currently support adding Templates to a project.
+
+- [Visual Studio Code Remote - Containers](../docs/specs/supporting-tools.md/#visual-studio-code-remote---containers)
+- [GitHub Codespaces](../docs/specs/supporting-tools.md/#github-codespaces)
+
+## Adding a Template to a project or codespace
+
```
Chuxel, line 76: Since this section is tool / service specific, I'm not sure it belongs in the spec as much as it does documentation for VS Code and Codespaces. It could be added to tool specific steps.  A CLI-based example would be better if there's a need to have something here.

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,192 @@
...
+| `optionId.type` | string | Type of the option. Valid types are currently: `boolean`, `string` |
+| `optionId.description` | string | Description for the option. |
+| `optionId.proposals` | array | A list of suggested string values. Free-form values **are** allowed. Omit when using `optionId.enum`. |
+| `optionId.enum` | array | A strict list of allowed string values. Free-form values are **not** allowed. Omit when using `optionId.proposals`. |
+| `optionId.default` | string | Default value for the option. |
+| `optionId.replaceIn` | array | List of file paths which the supporting tool should use to perform the string replacement of `optionId` with the selected value. The provided path is always relative to the sub-directory of the Template. |
+
+> `Note`: The `options` must be unique for every `devcontainer-template.json`
+
+### Template Supporting Tools and Services
```
Chuxel, line 68: As with the next section, I'm not sure we need to reference this as a part of the proposal as much as it should be in docs for VS Code and Codespaces.  We could, however, add this into the spec's "implementing tools" doc.  

You could refactor this to be "example usage" if you think it's important, but it might be better just to omit and add the implementing tools part when we move it into containers.dev.

proposals/devcontainer-templates.md
samruddhikhandale: Makes sense. Removed `type`

proposals/devcontainer-templates.md
samruddhikhandale: Agreed, added `publisher` 👍

proposals/devcontainer-templates.md
samruddhikhandale: Good catch, updated wordings.

proposals/devcontainer-templates.md
samruddhikhandale: Nice, updated!

proposals/devcontainer-templates.md
samruddhikhandale: Makes sense, removed!

proposals/devcontainer-templates.md
samruddhikhandale: Makes sense, removed. 👍

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+This specification defines a pattern where community members and organizations can author and self-publish [Dev Container Templates](./devcontainer-templates.md). 
+
+Goals include:
+
+- For Template authors, create a "self-service" way to publish a Template, either publicly or privately, that is not centrally controlled.
+- Provide the ability to standardize publishing such that supporting tools may implement their own mechanism to aid Template discoverability as they see fit.
+
+## Source code
+
+Templates source code is stored in a git repository.
```
bamurtaugh, line 12: Potential nit?
```suggestion
A Template's source code is stored in a git repository.
```

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+For ease of authorship and maintenance, [1..n] Templates can share a single git repository. This set of Templates is referred to as a "collection," and will share the same [`devcontainer-collection.json`](#devcontainer-collection.json) file and "namespace" (eg. `<owner>/<repo>`).
+
+> **Note:** Templates and [Features](./devcontainer-features.md) should be placed in different git repositories. 
+
+Source code for the set follows the example file structure below:
```
bamurtaugh, line 18:
```suggestion
Source code for a set of Templates follows the example file structure below:
```

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+### devcontainer-collection.json
+
+The `devcontainer-collection.json` is an auto-generated metadata file.
+
+| Property | Type | Description |
+| :--- | :--- | :--- |
+| `sourceInformation` | object | Metadata from the implementing packaging tool. |
+| `templates` | array | The list of Templates that are contained in this collection.|
+
+Each templates's `devcontainer-template.json` metadata file is appended into the `templates` top-level array.
```
bamurtaugh, line 72: Consistency and spelling nit:
```suggestion
Each Template's `devcontainer-template.json` metadata file is appended into the `templates` top-level array.
```

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Distribution
+
+There are several supported ways to distribute templates.  Distribution is handled by the implementing packaging tool.
```
bamurtaugh, line 76:
```suggestion
There are several supported ways to distribute Templates.  Distribution is handled by the implementing packaging tool.
```

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+A user can add the Templates in to their projects as defined by the [Templates Supporting Tools](./devcontainer-templates.md/#adding-a-template-to-a-project-or-codespace).
```
bamurtaugh, line 78:
```suggestion
A user can add Templates in to their projects as defined by the [Templates Supporting Tools](./devcontainer-templates.md/#adding-a-template-to-a-project-or-codespace).
```

Unless we're referring to a specific set of Templates (which would thus be "The" Templates?)

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+### OCI Registry
+
+An OCI registry that implements the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec) serves as the primary distribution mechanism for Templates.
+
+Each packaged Template is pushed to the registry following the naming convention `<registry>/<namespace>/<id>[:latest]`.
+
+> **Note:** The `namespace` is a unique identifier for the collection of templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`.
```
bamurtaugh, line 86:
```suggestion
> **Note:** The `namespace` is a unique identifier for the collection of Templates and must be different than the collection of [Features](./devcontainer-features.md). There are no strict rules for the `namespace`; however, one pattern is to set `namespace` equal to source repository's `<owner>/<repo>`. 
```

proposals/devcontainer-templates-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+ARTIFACT_PATH=devcontainer-template-go.tgz
+
+oras push ${REGISTRY}/${NAMESPACE}/${TEMPLATE}:latest \
+        --manifest-config /dev/null:application/vnd.devcontainers \
+                ./${ARTIFACT_PATH}:application/vnd.devcontainers.layer.v1+tar
+
+```
+
+The "namespace" is the globally identifiable name for the collection of templates. (eg: `owner/repo` for the source code's git repository).
```
bamurtaugh, line 108:
```suggestion
The "namespace" is the globally identifiable name for the collection of Templates. (eg: `owner/repo` for the source code's git repository).
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,177 @@
+# Dev Container Templates reference
+
+Development container "Templates" are source files packaged together that encode configuration for a complete development environment. A Template can be used in a new or existing project and a [supporting tool](https://containers.dev/supporting) will use the configuration from the template to build a development container.
```
bamurtaugh, line 3:
```suggestion
Development container "Templates" are source files packaged together that encode configuration for a complete development environment. A Template can be used in a new or existing project, and a [supporting tool](https://containers.dev/supporting) will use the configuration from the Template to build a development container.
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,177 @@
...
+
+The configuration is placed in a [`.devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) which can also reference other files within the Template. Alternatively, `.devcontainer/devcontainer.json` can also be used if the container needs to reference other files, such as a `Dockerfile` or `docker-compose.yml`. A template can also provide additional source files (eg: boilerplate code or a [lifecycle script](https://containers.dev/implementors/json_reference/#lifecycle-scripts).
```
bamurtaugh, line 5:
```suggestion
The configuration is placed in a [`.devcontainer.json`](/docs/specs/devcontainer-reference.md#devcontainerjson) which can also reference other files within the Template. Alternatively, `.devcontainer/devcontainer.json` can also be used if the container needs to reference other files, such as a `Dockerfile` or `docker-compose.yml`. A Template can also provide additional source files (eg: boilerplate code or a [lifecycle script](https://containers.dev/implementors/json_reference/#lifecycle-scripts).
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,177 @@
...
+
+The `id` format (`<oci-registry>/<namespace>/<template>[:latest]`) dictates how a [supporting tool](../docs/specs/supporting-tools.md#supporting-tools-and-services) will locate and download a given Template from an OCI registry (For example - `ghcr.io/user/repo/go:latest`). The registry must implement the [OCI Artifact Distribution Specification](https://github.com/opencontainers/distribution-spec). Some implementors can be [found here](https://oras.land/implementors/). 
+
+## Versioning
+
+Each Template is published with only the `latest` tag. Tooling that handles releasing Templates will republish the `latest` tag every time a new release is created.
+
+## Release
+
+_For information on distribution Templates, see [devcontainer-templates-distribution.md](./devcontainer-templates-distribution.md)._
```
bamurtaugh, line 78:
```suggestion
_For information on distributing Templates, see [devcontainer-templates-distribution.md](./devcontainer-templates-distribution.md)._
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Option Resolution
+
+A templates's `options` is used by a supporting tool to prompt for different configuration options. A supporting tool will parse the `options` object provided by the user. If a value is selected for a Template, it will be replaced in the provided `replaceIn` files.
```
bamurtaugh, line 82:
```suggestion
A Template's `options` property is used by a supporting tool to prompt for different configuration options. A supporting tool will parse the `options` object provided by the user. If a value is selected for a Template, it will be replaced in the provided `replaceIn` files.
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Option resolution example
+
+Consider a `java` template with the following folder structure:
```
bamurtaugh, line 86:
```suggestion
Consider a `java` Template with the following folder structure:
```

proposals/devcontainer-templates.md
```diff
@@ -0,0 +1,177 @@
...
+		"ghcr.io/devcontainers/features/node:1": {
+			"version": "${templateOption:nodeVersion}",
+      "installMaven": "${templateOption:installMaven}"
+		}
+	},
+//	...
+}
+```
+
+A user tries to add the `java` Template to their project using the [supporting tools](../docs/specs/supporting-tools.md#supporting-tools-and-services) and selects `17-bullseye` when prompted for `"Specify version of Go."` and uses the `default` values for `"Specify version of node, or 'none' to skip node installation."` and `"Install Maven, a management tool for Java."`.
```
bamurtaugh, line 149:
```suggestion
A user tries to add the `java` Template to their project using the [supporting tools](../docs/specs/supporting-tools.md#supporting-tools-and-services) and selects `17-bullseye` when prompted for `"Specify version of Go"` and the `default` values when prompted for `"Specify version of node, or 'none' to skip node installation"` and `"Install Maven, a management tool for Java"`.
```

bamurtaugh (commented):
Thanks for the great progress here!

proposals/devcontainer-templates-distribution.md
chrmarti: #22 would reference only a devcontainer.json as I understand it, not an entire template.

My take-away on template versioning was that it would not solve the schema version compatibility (the same way the package.json's version property doesn't solve the Node.js version compatibility) and that we would look into adding a `"schemaVersion"` (or similar) property (similar to the package.json's engines property) in a separate proposal to not overload this one.

proposals/devcontainer-templates.md
chrmarti: Sorry, I wasn't clear: I suggest to keep `type` and we can add tooling support later (whenever we get to it). :)

proposals/devcontainer-templates-distribution.md
Chuxel: I am quite confident that when we expand #22 to supporting remote devcontainer.jsons, customers will want an entire template since this is generally how the request has come across.  We limited #22 to local to start because this proposal was not in place. The request we've heard often is to have the ability to reference a remote template so that when the template updates with non-breaking changes, downstream repositories get the updates without modification. To achieve this, we'd need to support referencing MAJOR or MAJOR.MINOR versions of templates like we do with features.

Also, with the schema version property, I agree that we do not want to overload this proposal, but fundamentally that will require being able to go back to a previous version of a template.

In either case, including the same versioning semantics not only keeps things consistent with features, but lets clients implement as they see fit and we don't have to retroactively go back and update this proposal and how anyone that is already publishing templates is doing so.

We can omit this from the proposal here - but given what we already know, and how features is set up, why would we do that?

samruddhikhandale: Writing down my thoughts for semver vs always publishing `latest` (mostly reiterating what Chuck already mentioned)
- Using semantic versioning will make publishing consistent with Features - it will avoid questions from the community of why one logic for Features and other for Templates.
- Given current consumption of definitions/templates by the Remote Containers UX, I agree versioning does not make complete sense and it doesn't sound helpful. Hence, for UX we can always pin the `latest` tag. However, there is a possibility that the community could develop some tool or service which might require pinning specific versions. So it will be nice to have (just in case)
    - Given our current set of maintained Templates, we might not think that we may add breaking changes to them. However, given that the Templates will be published and distributed by community as well, we can't be sure what pattern they might use. 

To sum up, I believe that having semver may not have useful scenarios currently, but it could be nice to have to provide flexibility and adaptability for future scenarios. Also, I don't see a downside of publishing semantic versions, unless I am missing something super obvious 😄

chrmarti: I agree it might be useful in the future, I'm fine with adding it now.

chrmarti (approved):
LGTM!

Chuxel (approved):

eitsupi:
Hi, thanks for working on this.
Are you planning to create a template for this as well as feature?

samruddhikhandale:
Hi @eitsupi,
Yes, we will create a Template repo as well when self authoring/publishing mechanism is in place.

eitsupi:
Glad to hear it, and I look forward to seeing the devcontainer ecosystem expand even further!

PR #105, opened 2022-09-28
Template: Modify devcontainer-template.json properties

samruddhikhandale:


proposals/devcontainer-templates.md
```diff
@@ -47,22 +47,20 @@ The `options` property contains a map of option IDs and their related configurat
       "type": "string",
       "description": "Description of the option",
       "proposals": ["value1", "value2"],
-      "default": "value1",
-      "replaceIn": [".devcontainer.json"],
```
samruddhikhandale, line 51: Removed `replaceIn` as the tooling can find and replace in all the files within the Template folder.

proposals/devcontainer-templates.md
```diff
@@ -26,19 +26,19 @@ The properties of the file are as follows:
 | Property | Type | Description |
 | :--- | :--- | :--- |
 | `id` | string | ID of the Template. The `id` should be unique in the context of the repository/published package where the Template exists and must match the name of the directory where the `devcontainer-template.json` resides. |
+| `version` | string | The semantic version of the Template. |
 | `name` | string | Name of the Template. |
 | `description` | string | Description of the Template. |
 | `documentationURL` | string | Url that points to the documentation of the Template. |
 | `licenseURL` | string | Url that points to the license of the Template. |
-| `type` | string | Type of the dev container (image, dockerfile, dockerCompose) created by the Template. |
```
samruddhikhandale, line 33: CLI template package/publish commands will add the logic to determine it's value.

samruddhikhandale, line 33: @jkeech when `type=image`, we can not be sure if the dev container only consists of a single `.devcontainer.json` ([example](https://github.com/devcontainers/templates/tree/main/src/jekyll/.devcontainer)) which can be used by dotcom.

Do you recommend adding another property (say `isSingleConfigFile`: boolean) to the `devcontainer-template.json` (which will be automatically added/computed by the cli publish cmd) ?

jkeech: I think the publish command should inject something like `fileCount: number` when it's processing a feature. It shouldn't need to be manually entered into devcontainer-template.json. Doing it automatically would also prevent the declared value from getting out of sync with the template contents.

chrmarti (approved):
LGTM!

fuller53 (approved):

PR #108, opened 2022-10-03
Update schema with new properties (microsoft/vscode-remote-release#7278)

chrmarti:


schemas/devContainer.base.schema.json
```diff
@@ -164,6 +164,55 @@
...
+				"containerEnv": {
+					"type": "object",
+					"additionalProperties": {
+						"type": "string"
+					},
+					"description": "Container environment variables."
+				},
+				"containerUser": {
+					"type": "string",
+					"description": "The user the container will be started with. The default is the user on the Docker image."
```
samruddhikhandale, line 176: > The default is the user on the Docker image

Is it possible that the default user is set by a Feature?

samruddhikhandale (approved):
LGTM. Approving with a comment.

joshspicer (approved):

schemas/devContainer.base.schema.json
chrmarti: Not currently. (The containerUser and remoteUser from the devcontainer.json will be included in the image metadata. So these are passed on from the base images in devcontainers/images.)

PR #118, opened 2022-10-14
Templates proposal - Use `.devcontainer/devcontainer.json`

samruddhikhandale:
Use `.devcontainer/devcontainer.json` instead of `.devcontainer.json` - so that all the dev config files for templates can always be found under a single `.devcontainer` folder


joshspicer (approved):
FWIW i'm indifferent with guiding people toward one or the other, but this change looks good to me and is consistent with the team's consensus!

PR #122, opened 2022-10-20
Add new properties and label support to reference

Chuxel:
This updates the devcontainer.json reference per https://github.com/devcontainers/spec/blob/main/proposals/image-metadata.md

The main spec document and the merge logic in the proposal still needs to be updated in the main spec.

It uses an emoji to indicate which properties are supported in an image label, but given most properties are supported, we could consider flipping this to indicate which are not supported instead.


docs/specs/devcontainerjson-reference.md
```diff
@@ -1,24 +1,33 @@
-# devcontainer.json reference
+# Dev Container metadata reference
 
 The `devcontainer.json` file contains any needed metadata and settings required to configurate a **development container** for a given well-defined tool and runtime stack. It can be used by [tools and services that support the dev container spec](supporting-tools.md) to create a **development environment** that contains one or more **development containers**.
 
+Metadata properties marked with a 🏷 can also be added to an **image label** in addition to devcontainer.json file. The `devcontainer.metadata` image label can contain an array of json snippets that will be automatically merged with devcontainer.json contents (if any) when a container is created.
```
bamurtaugh, line 5: Small nit

```suggestion
Metadata properties marked with a 🏷 can also be added to an **image label** in addition to a devcontainer.json file. The `devcontainer.metadata` image label can contain an array of json snippets that will be automatically merged with devcontainer.json contents (if any) when a container is created.
```

bamurtaugh (approved):
LGTM, thanks so much for opening this up! I think adding the emoji to supported properties makes sense - although it is most of them, it's a nice way to be able to scan through and see if you can add a property to a label. And I'm not sure we'd have as nice of an emoji option to indicate no label support :)

Chuxel (commented):
Language tweak

docs/specs/devcontainerjson-reference.md
Chuxel: I made an update in a subsequent commit - see what you think of the updated language.

bamurtaugh: LGTM! I like the updated flow.

Issue #82, opened 2022-08-11
Improve hostRequirements, Features support for dev containers that use GPUs


Chuxel:
Dev containers can run on machines with GPUs and access them directly on Linux or through WSL on Windows. However, currently there is no direct support for a GPUs in the spec.

To enable GPU support, I'd propose we enhance the "hostRequirements" property to allow specifying GPU information.  If this information is present, we then automatically enable GPU support (e.g., via --gpus all).

I'd propose we support a form that simply adds a requirement for any GPU, with the option to specify more requirements for further validation. Here's a proposed format:

Just enabling/requiring any GPU

```jsonc
"hostRequirements": {
    "gpu": true
```

Specifying cores and memory:
```jsonc
"hostRequirements": {
    "gpu": {
        "cores": 2
        "memory": "4gb" // Same value format as memory for the machine
    }
}
```

and with type

```
"hostRequirements": {
    "gpu": {
        "cores": 2
        "memory": "4gb"
        "type": "nvidia"
    }
}
```

To allow Dev Container Features (#61) to require a GPU (e.g., in an NVIDIA CUDA feature), I'd propose we also allow `hostRequirements` to be specified in `devcontainer-features.json`.

chris-major-improbable:
I generally like this - however I'd really like to be able to have a devcontainer that supports GPUs when available but doesn't require them. 

For my specific use case, we have some libraries that can use GPUs, but also fall back to CPUs if they are not available. Currently the developers need to manually add/remove the `-gpus all`.

My sugestion would be to extended this proposal with something like:
```json
"hostRequirements": {
    "gpu": {
        "optional": true
    }
}
```
(I appreciate that mixing 'requirements' and `optional` elements seems a bit counter intuitive, however I can't think of a better option).

Clockwork-Muse:
> we then automatically enable GPU support (e.g., via --gpus all).

This flag isn't present in podman, and isn't part of the CDI spec, btw.  [You're supposed to set things up differently in podman](https://github.com/containers/podman/pull/12825) (although you can also install a global hook).

chrmarti:
Wrote down what we agreed on for the implementation here: https://github.com/devcontainers/spec/issues/82

PR #143, opened 2022-11-28
GPU host requirement proposal (devcontainers/spec#82)

chrmarti:


Chuxel (approved):

l0rem1psum:
Based on my current workflow and use cases, I feel that it will also be useful to include GPU architecture as one of the requirement fields. In the case of NVIDIA GPUs, for instance, we can indicate whether T4 or P4 GPUs. It will also be useful to indicate the number of GPUs required.

chrmarti:
@l0rem1psum Number of different GPUs? (We support number of cores.) It might make sense to first collect feedback from the use of the currently proposed approach and then extend it to multiple GPUs and specific GPU types. (Although "GPU types" sounds like a constantly changing set of identifiers and capabilities which we might not want to track in a specification.)

l0rem1psum:
There might be applications where multiple GPUs of different architectures/models are required, but I have only worked with multiple GPUs of the same configuration. On a side note, using the term "cores" on GPUs may oftentimes cause confusion as far as I understand since a single GPU has hundreds to thousands of cores.

l0rem1psum:
Not sure about other users, the few requirements I often care about are as follow:

1. Number of GPU units
2. GPU architecture
3. GPU memory
4. GPU compute capability 
5. GPU driver version
6. CUDA version (not really a host requirement but is bound by driver version)

joshspicer (approved):
This sounds reasonable to me.  The additional comments on this PR indicate additional properties that would be useful that I don't have GPU experience to speak to those comments

proposals/gpu-host-requirement.md
```diff
@@ -0,0 +1,23 @@
+# GPU host requirement
+
+## Goal
+
+This proposal adds support for GPU to the existing support for host requirements.
```
bamurtaugh, line 5: Small wording suggestion:

```suggestion
This proposal adds GPU support to the existing host requirements properties.
```

proposals/gpu-host-requirement.md
```diff
@@ -0,0 +1,23 @@
...
+- Community PR: https://github.com/devcontainers/cli/pull/173
+
+## Proposal
+
+We propose to add a new `gpu` property to the `hostRequirements` object in the devcontainer.json schema. The property can be a boolean value, the string `optional` or an object:
+
+(Additional row for the existing table in the spec.)
+| Property | Type | Description |
+|----------|------|-------------|
+| `hostRequirements.gpu` 🏷️ | boolean \| 'optional' | Indicates whether or not a GPU is required. A value 'optional' indicates that might be used if available. The object is described in the table below. The default is false. For example: `"hostRequirements": "optional"` |
```
bamurtaugh, line 16:
```suggestion
| `hostRequirements.gpu` 🏷️ | boolean | 'optional' | Indicates whether or not a GPU is required. A value 'optional' indicates that might be used if available. The object is described in the table below. The default is false. For example: `"hostRequirements": "optional"` |
```

chrmarti: The backslash is needed to render the pipe symbol. Without the backslash the pipe separates two columns. Thanks.

chrmarti:
Thanks @l0rem1psum, I have opened a new issue to track your suggestions.

PR #148, opened 2022-12-02
Update `remoteUser` info

bamurtaugh:
To address the outdated reference mentioned [here](https://github.com/devcontainers/templates/pull/65#issuecomment-1324652160).

Next steps:
- [ ] Reach alignment on this PR (placement, wording)
- [ ] Merge in same changes for containers.dev - started PR: https://github.com/devcontainers/devcontainers.github.io/pull/93
- [ ] Update https://aka.ms/dev-containers-non-root to point to new content in containers.dev (easier to traverse as end user rather than diving deep into spec repo itself)


docs/specs/devcontainerjson-reference.md
```diff
@@ -167,3 +167,9 @@ You can see the VS Code implementation of the dev container schema [here](https:
 ## Publishing vs forwarding ports
 
 Docker has the concept of "publishing" ports when the container is created. Published ports behave very much like ports you make available to your local network. If your application only accepts calls from `localhost`, it will reject connections from published ports just as your local machine would for network calls. Forwarded ports, on the other hand, actually look like `localhost` to the application.
+
+## remoteUser
+
+A dev container configuration will inherit the `remoteUser` property from the base image it uses.
+
+Using the [images](https://github.com/devcontainers/images) and [Templates](https://github.com/devcontainers/images) part of the spec as an example: `remoteUser` in these images is set to a custom value of `vscode` - you may view an example in the [C++ image](https://github.com/devcontainers/images/blob/main/src/cpp/.devcontainer/devcontainer.json#L26). The [C++ Template](https://github.com/devcontainers/images/blob/main/src/cpp/.devcontainer/devcontainer.json#L26) will then inherit the custom `vscode` user from [its base C++ image](https://github.com/devcontainers/templates/blob/main/src/cpp/.devcontainer/Dockerfile#L1).
```
samruddhikhandale, line 175: ℹ️ Not all images in devcontainers/images have `vscode` as remoteUser. `typescript-node` & `javascript-node` have `remoteUser:node` and `universal` has `remoteUser: codespace`.

Wondering if this statement might be a lil confusing? 🤔 
Maybe we  shouldn't generalize this example for all images?

bamurtaugh: Ah yeah I think that's a great point. Updated to be a bit more general - let me know what you think.

docs/specs/devcontainerjson-reference.md
```diff
@@ -167,3 +167,9 @@ You can see the VS Code implementation of the dev container schema [here](https:
 ## Publishing vs forwarding ports
 
 Docker has the concept of "publishing" ports when the container is created. Published ports behave very much like ports you make available to your local network. If your application only accepts calls from `localhost`, it will reject connections from published ports just as your local machine would for network calls. Forwarded ports, on the other hand, actually look like `localhost` to the application.
+
+## remoteUser
+
+A dev container configuration will inherit the `remoteUser` property from the base image it uses.
+
+Using the [images](https://github.com/devcontainers/images) and [Templates](https://github.com/devcontainers/images) part of the spec as an example: `remoteUser` in these images is set to a custom value - you may view an example in the [C++ image](https://github.com/devcontainers/images/blob/main/src/cpp/.devcontainer/devcontainer.json#L26). The [C++ Template](https://github.com/devcontainers/images/blob/main/src/cpp/.devcontainer/devcontainer.json#L26) will then inherit the custom `remoteUser` value from [its base C++ image](https://github.com/devcontainers/templates/blob/main/src/cpp/.devcontainer/Dockerfile#L1).
```
samruddhikhandale, line 175:
```suggestion
Using the [images](https://github.com/devcontainers/images) and [Templates](https://github.com/devcontainers/templates) part of the spec as an example: `remoteUser` in these images is set to a custom value - you may view an example in the [C++ image](https://github.com/devcontainers/images/blob/main/src/cpp/.devcontainer/devcontainer.json#L26). The [C++ Template](https://github.com/devcontainers/templates/tree/main/src/cpp) will then inherit the custom `remoteUser` value from [its base C++ image](https://github.com/devcontainers/templates/blob/main/src/cpp/.devcontainer/Dockerfile#L1).
```

Nit: fixing links

docs/specs/devcontainerjson-reference.md
samruddhikhandale: Nice, like the new wordings.

docs/specs/devcontainerjson-reference.md
bamurtaugh: Ah oops! Thank you!

samruddhikhandale (approved):
Looks good! thanks. ✨ 

ℹ️  The cpp example in this PR is great and makes a clear point. However, as this PR links to specific lines in the images/templates repo, it is very susceptible to breaking. 🤔 

I am good to approve, but maybe we need to be update URLs if cpp image/template updates (which should be rare I think)?

bamurtaugh:
Thanks for the quick review and great help, @samruddhikhandale!

Makes sense on breakages. Let's go ahead and merge since I agree it's nice to have a concrete example, and we'll just be aware that we'll need to be cautious if there are major changes. An alternative is to just link to the whole file rather than specific lines, but I think it's really helpful to point the user to the specific lines.

chrmarti:
LGTM, thanks!

PR #185, opened 2023-01-20
Create devcontainer.json

Mohammedalanizi92:


samruddhikhandale (commented):
Thanks for putting up this PR. However, this repo is meant for documentation & doesn't need the proposed changes which add the tooling.

Issue #181, opened 2023-01-17
[proposal] Allow Features to contribute lifecycle hooks


joshspicer:
relevant to: https://github.com/devcontainers/spec/issues/60

**Proposal**

 Introduce the following properties to [devcontainer-feature.json](https://containers.dev/implementors/features/#devcontainer-feature-json-properties), mirroring the behavior and syntax of the `devcontainer.json` lifecycle hooks.

- "onCreateCommand"
- "updateContentCommand"
- "postCreateCommand"
- "postStartCommand"
- "postAttachCommand"

## Example

```jsonc
{
   "id": "my-feature",
   "version": "1.0.0"
   "postCreateCommand": "./myScript.sh"
}

```

Lifecycle scripts are looked up relative to the root of the Features source directory, which is bundled up as an archive when distributed.  Features already support distributing arbitrary scripts via the registry, this proposal will now allow the Feature to contribute to the build's lifecycle system.  

Feature lifecycle hooks will be prepended to the lifecycle hooks (if any) declared by the current `devcontainer.json`. Lifecycle hooks will be prepended in the order that the Feature's installation script was executed.  Commands will be joined in such a way that a failure in any one lifecycle hook, will indicate a failure for that entire lifecycle hook execution.

bamurtaugh:
Should we label this with finalization, now that the proposal is merged into the spec?

eitsupi:
Hi, I want use this functionality to JuliaLang/devcontainer-features#9

Is there a recommended path for scripts to be placed in containers at this time?
For example, at the moment this is placed on a path like `/julia-devcontainer-scripts/postcreate.jl`, which I feel is not a very good path.
https://github.com/julia-vscode/julia-devcontainer/blob/60c2852703d5f4e2fc419198a95c49be37d12ccb/Dockerfile#L33

It would be great if the recommended paths were mentioned in the documentation.

joshspicer:
> Is there a recommended path for scripts to be placed in containers at this time?

Given that any base image can be used with Features, I don't think our guidance on _where_ to place scripts would be too meaningful. Functionally the script needs to be readable and executable by the remoteUser

Looking at https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard, somewhere like `/usr/share/devcontainer/feature` might be a natural spot.  In the first cut of this feature I was thinking of persisting the entire Feature source in that folder, and may use that (or similar) in the future if there's a reason to move from `/tmp`.

[...]

PR #196, opened 2023-01-30
Add proposal for features contributing lifecycle scripts

joshspicer:
ref: https://github.com/devcontainers/spec/issues/60, https://github.com/devcontainers/spec/issues/181

### DIFF
--- a/proposals/features-contribute-lifecycle-scripts.md
+++ b/proposals/features-contribute-lifecycle-scripts.md
@@ -18,50 +18,52 @@ Allow Feature authors to provide lifecycle hooks for their Features during a dev
 
 Note that `initializeCommand` is omitted, pending further discussions around a secure design.
 
-Additionally, introduce a `${featureRoot}` environment variable, which is expanded at build time to the root of the expanded Feature the variable is used from.  
+Additionally, introduce a `${featureRoot}` dev container variable, which is expanded at build time to the root of the Feature the variable is referenced from.  
 
-Feature lifecycle hooks will be prepended to the lifecycle hooks (if any) declared by the current `devcontainer.json`. Lifecycle hooks will be prepended in the order that the Feature's installation script was executed.  Commands will be joined in such a way that a failure in any one lifecycle hook, will indicate a failure for that entire lifecycle hook execution.
+As with all lifecycle hooks, commands are executed from the context (cwd) of the [project workspace folder](https://containers.dev/implementors/spec/#project-workspace-folder).
 
-### Parallel Execution
+All other semantic match the existing [Lifecycle Scripts](https://containers.dev/implementors/json_reference/#lifecycle-scripts) behavior exactly.
 
-Any Features that declare lifecycle hooks using the [parallel execution syntax](https://containers.dev/implementors/spec/#parallel-exec) will be executed in parallel with any other Features or user-contributed lifecycle scripts. Any Features that do not use this syntax will continue to be run in sequence, blocking subsequent runs until the script exits successfully.
+###  Execution
 
-Each parallel stage inherited from a Feature will be prefixed with the Feature `id`. Eg: `featureA_migrateDB`.
+Any Features that declare one of the aforementioned lifecycle hook properties will have their command executed _in parallel with any other Features, or user-contributed, lifecycle scripts_ during the target lifecycle point.
+
+> A consequence of executing scripts in parallel requires that operations in one script  do not block another. The execution order of individual lifecycle scripts within a given lifecycle hook is undefined, and any perceived ordering should not be relied upon.
+
+See here for more information on [lifecycle script parallel execution](https://containers.dev/implementors/spec/#parallel-exec).
 
 ## Examples
 
 
 ### Executing a script in the workspace folder
 
-The follow example illustrates executing a `postCreateCommand` script in the [project workspace folder](https://containers.dev/implementors/spec/#project-workspace-folder).  
+The follow example illustrates contributing an `onCreateCommand` and `postCreateCommand` script to `featureA`.  At each lifecycle hook during the build, the provided script will be executed in the [project workspace folder](https://containers.dev/implementors/spec/#project-workspace-folder), following the same semantics as [user-defined lifecycle hooks](https://containers.dev/implementors/json_reference/#lifecycle-scripts).
 
 ```jsonc
 {
    "id": "featureA",
-   "version": "1.0.0"
-   "postCreateCommand": "scriptInWorkspaceFolder.sh"
+   "version": "1.0.0",
+   "onCreateCommand": "myOnCreate.sh && myOnCreate2.sh",
+   "postCreateCommand": "myPostCreate.sh"
 }
 
 ```
 
 ### Executing a script bundled with the Feature
 
-The follow example illustrates executing a `postCreateCommand` script bundled with the Feature by using the `${featureRoot}` variable.
+The following example illustrates executing a `postCreateCommand` script bundled with the Feature utilizing the `${featureRoot}` variable.
 
 ```jsonc
 {
    "id": "featureB",
-   "version": "1.0.0"
-   "postCreateCommand": "${featureRoot}/bundledScript.sh",
-   "installsAfter": [
-        "featureA"
-   ]
+   "version": "1.0.0",
+   "postCreateCommand": "${featureRoot}/bundledScript.sh"
 }
 ```
 
-At build time, the `${featureRoot}` variable will be expanded to the temporary directory that contains all the Feature's assets.  For example, it may be expanded to `/tmp/vsch/container-features/featureB_1`.
+At build time, the `${featureRoot}` variable will be expanded to the temporary directory within the container that contains all the Feature's assets.  For example, it may be expanded to `/tmp/vsch/container-features/featureB_1`.
 
-When authored, `featureB`'s file structure would look something like:
+Given this example, `featureB`'s file structure would look something like:
 
 ```
 .
@@ -72,9 +74,9 @@ When authored, `featureB`'s file structure would look something like:
 │   │   └── install.sh
 ```
 
-### Installation Order (without parallel execution syntax)
+### Installation
 
-Given the following `devcontainer.json` and the two examples Features above.
+Given the following `devcontainer.json` and the two example Features above.
 
 ```jsonc
 {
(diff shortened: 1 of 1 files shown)


proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+
+Note that `initializeCommand` is omitted, pending further discussions around a secure design.
+
+Additionally, introduce a `${featureRoot}` environment variable, which is expanded at build time to the root of the expanded Feature the variable is used from.  
+
+Feature lifecycle hooks will be prepended to the lifecycle hooks (if any) declared by the current `devcontainer.json`. Lifecycle hooks will be prepended in the order that the Feature's installation script was executed.  Commands will be joined in such a way that a failure in any one lifecycle hook, will indicate a failure for that entire lifecycle hook execution.
+
+### Parallel Execution
+
+Any Features that declare lifecycle hooks using the [parallel execution syntax](https://containers.dev/implementors/spec/#parallel-exec) will be executed in parallel with any other Features or user-contributed lifecycle scripts. Any Features that do not use this syntax will continue to be run in sequence, blocking subsequent runs until the script exits successfully.
```
chrmarti, line 27: Currently all scripts of a lifecycle hook are run in parallel. We can change that, but it seems a bit artificial to make this dependent on the syntax used.

The different hooks run sequentially, e.g., any "postStartCommand" script will only be run after all "postCreateCommand" scripts have finished. We probably want to keep it this way, the wording in the proposal should make this clearer. (The example in the linked doc needs to be updated to run the commands in the background, these "postCreateCommand" scripts will block later hooks from running.)

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+- "updateContentCommand"
+- "postCreateCommand"
+- "postStartCommand"
+- "postAttachCommand"
+
+Note that `initializeCommand` is omitted, pending further discussions around a secure design.
+
+Additionally, introduce a `${featureRoot}` environment variable, which is expanded at build time to the root of the expanded Feature the variable is used from.  
+
+Feature lifecycle hooks will be prepended to the lifecycle hooks (if any) declared by the current `devcontainer.json`. Lifecycle hooks will be prepended in the order that the Feature's installation script was executed.  Commands will be joined in such a way that a failure in any one lifecycle hook, will indicate a failure for that entire lifecycle hook execution.
```
chrmarti, line 23: Do we expect the scripts to depend on each other, so we have to run them in installation order?

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Goal
+
+Allow Feature authors to provide lifecycle hooks for their Features during a dev container build.
+
+## Proposal
+
+ Introduce the following properties to [devcontainer-feature.json](https://containers.dev/implementors/features/#devcontainer-feature-json-properties), mirroring the behavior and syntax of the `devcontainer.json` lifecycle hooks.
+
+- "onCreateCommand"
```
chrmarti, line 13: Mention that the value can be a string or an array of strings. The string variation is run with `/bin/sh -c ...` whereas the string array variation is executed directly. (This is also the case in the devcontainer.json.)

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+- "updateContentCommand"
+- "postCreateCommand"
+- "postStartCommand"
+- "postAttachCommand"
+
+Note that `initializeCommand` is omitted, pending further discussions around a secure design.
+
+Additionally, introduce a `${featureRoot}` environment variable, which is expanded at build time to the root of the expanded Feature the variable is used from.
```
chrmarti, line 21: Also mention here or above that the cwd is the workspace folder. (I see it mentioned in the examples, but it makes sense to have it here too.)

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+   "id": "featureA",
+   "version": "1.0.0"
+   "postCreateCommand": "scriptInWorkspaceFolder.sh"
+}
+
+```
+
+### Executing a script bundled with the Feature
+
+The follow example illustrates executing a `postCreateCommand` script bundled with the Feature by using the `${featureRoot}` variable.
```
chrmarti, line 49: Typo: "following"

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+
+Given a Feature defined as follows:
+
+```jsonc
+{
+    "id": "featureC",
+    "version": "1.0.0",
+    "postCreateCommand": {
+        "server": "npm start",
+        "db": ["mysql", "-u", "root", "-p", "my database"]
```
chrmarti, line 108: If these commands don't exit, they prevent "postStartCommand" from running.

chrmarti (changes requested):
👍 Left a few comments.

proposals/features-contribute-lifecycle-scripts.md
joshspicer: No I don't think we should expect them to depend on each other, so I suppose it makes more sense to keep execution order more vague in order to _not_ have folks eventually depend on this behavior.

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Executing a script bundled with the Feature
+
+The follow example illustrates executing a `postCreateCommand` script bundled with the Feature by using the `${featureRoot}` variable.
+
+```jsonc
+{
+   "id": "featureB",
+   "version": "1.0.0"
+   "postCreateCommand": "${featureRoot}/bundledScript.sh",
```
samruddhikhandale, line 55: Would `featureRoot ` essentially be the `id`? or is it some path like xx/xx/${id}? 

Can you give an example.

samruddhikhandale: Ah, you already mention it in next line 👍

samruddhikhandale (commented):
Besides Christof's comment, everything else looks good to me.

proposals/features-contribute-lifecycle-scripts.md
joshspicer: All lifecycle hooks are run in parallel regardless of if the "parallel execution syntax" is used? That actually seems to line up with this [unit test](https://github.com/devcontainers/cli/compare/joshspicer/features-lifecycles-hooks#diff-05813acb3066795d8650780efe8bf09807c62a1cde94cf9d6b7ee0c67bdb73fbR48-R64) I wrote, where the execution of each script is not deterministic (which would make sense if they're running at the exact same time).

This was all documented based on my previous understanding of how this code worked - but I agree that ordering would be artificial and not something we should introduce with this change.

proposals/features-contribute-lifecycle-scripts.md
joshspicer: Yes, good call out.

proposals/features-contribute-lifecycle-scripts.md
joshspicer: Yep!  In practice I think it would look a lot like `/tmp/vsch/container-features/featureB_1` on linux machines, but that would be an implementation detail and mostly transparent to the user (unless they looked at the verbose logging)

proposals/features-contribute-lifecycle-scripts.md
joshspicer: I will collapse this spec down to indicate that scripts will indeed run in parallel.

joshspicer: Although, I can think of instances where running _all_ lifecycle hooks in parallel may cause inconsistent behavior.  

For example, say two Features provide an `onCreateCommand` that both try to install an `apt` package.  The lock file will be taken by one script, and the other will fail to run. This behavior also has the change of being inconsistent, which we want to avoid with dev containers as much as possible.

Since our current offering for running commands in parallel is clearly opt-in (using the object syntax), there's less concern that users will accidentally get themselves stuck in a spot like this.  Additionally, Features obscure the contents of a script, so folks may not even realize there is a race condition in two Features' lifecycle scripts.

Does that make sense? I want us to discuss that before I move forward with serialization/parallelization of lifecycle hooks.

markphip: Speaking as a feature author, I think my expectation would be that the scripts run in sequence, especially if the `installAfter` option was used. This is mainly because I think of the features as installing in a sequence, and so the scripts running in a similar sequence just seems like how it would work.

That said, running them in parallel feels fine and manageable. If I really need something else to finish first, it does not seem impossible to have the script I provided detect and wait for some condition to be satisfied .. and if I control the feature I need to wait for, I can even create some kind of marker file to make this easy.

chrmarti: I'm fine with running all in parallel for now, but agree with the above that it would make sense to run scripts from different features in sequence to avoid conflicts. Scripts from the same feature could still run in parallel as we could expect each feature to manage dependencies among its scripts. If we run the scripts in sequence, it will make sense to do it in the same sequence as we ran the install scripts in case the hooks also have the same dependencies.

bamurtaugh: I agree with what @chrmarti mentions above. 👍

Cindys82 (commented):
Can someone please give me some advice for what I need to be do to get this app to work for me please and thanks

bamurtaugh:
Hi @Cindys82, could you let us know what you're trying to get working (i.e. which app), and if you have any specific issues you can share? Thanks so much!

proposals/features-contribute-lifecycle-scripts.md
joshspicer: Mentioned the semantics are the same as previously lifecycle scripts (with a link).

Gesisfm:
https://apps.apple.com/tw/app/%E6%98%9F%E5%9F%8Eonline-%E6%98%9F%E5%B9%A3%E7%88%AD%E9%9C%B8%E6%88%B0%E6%90%B6%E5%84%84%E8%90%AC%E5%BD%A9%E9%87%91/id919043762

proposals/features-contribute-lifecycle-scripts.md
```diff
@@ -0,0 +1,114 @@
...
+
+- "onCreateCommand"
+- "updateContentCommand"
+- "postCreateCommand"
+- "postStartCommand"
+- "postAttachCommand"
+
+Note that `initializeCommand` is omitted, pending further discussions around a secure design.
+
+Additionally, introduce a `${featureRoot}` dev container variable, which is expanded at build time to the root of the Feature the variable is referenced from.
```
chrmarti, line 21: Would `${featureRootFolder}` be clearer and align better with the existing variables?

chrmarti (changes requested):
Left a comment on naming. LGTM otherwise.

proposals/features-contribute-lifecycle-scripts.md
joshspicer: Updating 👍

alexdima (approved):
👍

PR #206, opened 2023-02-28
Joshspicer/consolidate merged lifecycle hook schema

joshspicer:


AlainGama (approved):

AlainGama (approved):

chrmarti (changes requested):
Additionally:
- `customizations` is a map of objects in the devcontainer.json and a map of object arrays in the merged config.
- `entrypoints` is not in the devcontainer.json, but is an array of strings in the merged config.

How would we represent the origins of the lifecycle commands, so we can log them?

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,126 @@
...
+
+Implementing tools should follow the [documented merging logic](https://containers.dev/implementors/spec/#merge-logic) and output a `devcontainer.json` that represents the resolved configuration.
+
+There are a couple special cases to the merging logic that should be noted:
+
+#### B1. The `customizations` property
+
+The merging of the `customizations` property is left to the implementing tool, therefore there is no defined merging pattern. 
+
+A property `$customizations` should be added to the merged configuration as an array of customizations within each top-level customization object.  Tooling that reads the mergedConfiguration can process this property as needed.
```
chrmarti, line 69: Why not use a similar approach to the lifecycle commands and support object and array of object values in the `customizations` map's values including `$origin` properties?

chrmarti (changes requested):

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,126 @@
+# Goal
+
+This proposal aims to raise a concept of the 'mergedConfiguration' into the specification, so that implementing tools can generate, as well as consume, a 'merged' `devcontainer.json` file that is the computed merge after parsing a user 'devcontainer.json'.  [The merging logic is already documented](https://containers.dev/implementors/spec/#merge-logic). This proposal focuses on standarding an output JSON format to represent this merge.
```
bamurtaugh, line 3:
```suggestion
This proposal aims to raise a concept of the 'mergedConfiguration' into the specification, so that implementing tools can generate, as well as consume, a 'merged' `devcontainer.json` file that is the computed merge after parsing a user 'devcontainer.json'.  [The merging logic is already documented](https://containers.dev/implementors/spec/#merge-logic). This proposal focuses on standardizing an output JSON format to represent this merge.
```

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,126 @@
...
+
+Properties that can not or should not be represented as `devcontainer.json` properties are prepended with a `$`.  This proposal standardizes how these properies should be processed and outputted to file, so that implementing tools can write/consume them consistently.
```
bamurtaugh, line 5:
```suggestion
Properties that can not or should not be represented as `devcontainer.json` properties are prepended with a `$`.  This proposal standardizes how these properties should be processed and outputted to file, so that implementing tools can write/consume them consistently.
```

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,126 @@
...
+	]
+},
+...
+```
+
+### B2. The `entrypoint` property
+
+Dev Container Features are able to contribute an `entrypoint` property. This property is not available in the `devcontainer.json`.
+
+A property `$entrypoints` should be added to the merged configuration containing an array of entrypoint strings.  Tooling that reads the mergedConfiguration can process this property as needed.
```
bamurtaugh, line 112:
```suggestion
A property `$entrypoints` should be added to the merged configuration containing an array of entrypoint strings.  Tooling that reads the 'mergedConfiguration' can process this property as needed.
```

proposals/consolidate-merged-lifecycle-hook-schema.md
joshspicer: Thinking about it more I agree that makes sense. Updatesd

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,141 @@
...
+    // ...other devcontainer.json properties...
+	onCreateCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	updateContentCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	postCreateCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	postStartCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	postAttachCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+}
+```
+
+An optional parameter `$origin` is added to the `LifecycleCommandParallel` that supporting tools can use to indicate the source of the command.  For example, this is useful for outputting in the creation log which Feature provided a certain lifecycle hook.  The `$` notation is used to indicate this property is additional tooling metadata that should not be present in a user `devcontainer.json`.
```
chrmarti, line 36: Would `$origin` be the qualified feature id with the version OR `devcontainer.json`?

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,141 @@
...
+	]
+}
+...
+```
+
+### B. Generating a merged configuration
+
+The [reference implementation](https://github.com/devcontainers/cli) includes a 'read-configuration' command that can be used to resolve a fully merged configuration from a given `devcontainer.json`.
+
+Implementing tools should follow the [documented merging logic](https://containers.dev/implementors/spec/#merge-logic) and output a `devcontainer.json` that represents the resolved configuration.
```
chrmarti, line 119: Maybe omit this paragraph ("Implementing tools should...") as it may not make sense for every implementing tool to offer such a command (implementers might not write a CLI at all).

chrmarti (changes requested):

avidal:
Is there still an appetite for this? In our SDK/CLI I'm tempted to implement something similar to help wrangle types (ours is written in Go, so I unfortunately can't leverage such rich types to keep everything in one place).

Something that would help additionally would be some test cases for how to merge configuration that implementing tools can run against, but that's a matter for another time I think.

joshspicer:
Thanks for the nudge @avidal, this is definitely something I still want us to move forward with. I'll give this spec's comments another pass right now

proposals/consolidate-merged-lifecycle-hook-schema.md
joshspicer: Now that we have Feature dependencies and may have more than one Feature contributing configuration, we should use the "canonical ID" (for OCI Features, expanded with the SHA hash).

joshspicer: Although i'm going back and forth here.  At the end of the day, the intention I have for `$origin` is to make it easier to understand from the logs what is happening. It may be more intuitive to print verbatim what is in the `devcontainer.json`.

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,141 @@
...
+### B1. The `entrypoint` property
+
+Dev Container Features are able to contribute an `entrypoint` property. This property is not available in the `devcontainer.json`.
+
+A property `$entrypoints` should be added to the merged configuration containing an array of entrypoint strings.  Tooling that reads the 'mergedConfiguration' can process this property as needed.
+
+In the below example, the `entrypoint` property is contributed from two different sources.  The merged `$entrypoints` property would be an array of two strings, one for each entrypoint.
+
+```json
+"$entrypoints": [
```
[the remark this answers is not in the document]
joshspicer: I think this is ok.  I don't think we'll necessarily do anything with this value if it's in a `devcontainer.json` as input.  I imagine all the `$` are ignored and used by humans as helpful output metadata.

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,142 @@
...
+    // ...other devcontainer.json properties...
+	onCreateCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	updateContentCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	postCreateCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	postStartCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+	postAttachCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
+}
+```
+
+An optional parameter `$origin` is added to the `LifecycleCommandParallel` that supporting tools can use to indicate the source of the command.  For example, this is useful for outputting in the creation log which Feature provided a certain lifecycle hook.  The `$` notation is used to indicate this property is additional tooling metadata that should not be present in a user `devcontainer.json`.
```
avidal, line 37: What is the `$origin` of a bit of config that comes from the `devcontainer.metadata` image label (assuming it doesn't carry one already)?

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,142 @@
...
+
+The [reference implementation](https://github.com/devcontainers/cli) includes a 'read-configuration' command that can be used to resolve a fully merged configuration from a given `devcontainer.json`.
+
+Special cases are detailed below.
+
+### B1. The `entrypoint` property
+
+Dev Container Features are able to contribute an `entrypoint` property. This property is not available in the `devcontainer.json`.
+
+A property `$entrypoints` should be added to the merged configuration containing an array of entrypoint strings.  Tooling that reads the 'mergedConfiguration' can process this property as needed.
```
avidal, line 121: This paragraph says they are an "array of entrypoint strings" but the example json below shows them as an array of maps with an `entrypoint` key and an optional `$origin` key.

proposals/consolidate-merged-lifecycle-hook-schema.md
```diff
@@ -0,0 +1,142 @@
...
+LifecycleCommand = string | string[]
+LifecycleCommandParallel = { [key: string]: LifecycleCommand; $origin?: string };
+```
+
+In a `devcontainer.json`, each of the following lifecycle hooks shown below can now be represented by one of these three types.
+
+```typescript
+{
+    // ...other devcontainer.json properties...
+	onCreateCommand?: LifecycleCommand | LifecycleCommandParallel | LifecycleCommandParallel[]
```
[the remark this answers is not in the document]
avidal: Oh, this change specifically affects the schema for `devcontainer.json`, I see. I hadn't realized that `devcontainer-feature.json` has a distinct json schema.

proposals/consolidate-merged-lifecycle-hook-schema.md
joshspicer: If a section of metadata in `devcontainer.metadata` doesn't already have an `$origin`, we wouldn't anything.  It had to come from somewhere (either a base dev container that defined the image, a Feature, or the user's current dev container).  If omitted, we don't have any insight into where it came from and we'd need to log something generically in those cases.

proposals/consolidate-merged-lifecycle-hook-schema.md
joshspicer: Thanks, I edited the example yesterday and missed this sentence.

proposals/consolidate-merged-lifecycle-hook-schema.md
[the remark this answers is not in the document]
joshspicer: Yes, I don't think we'd want to update the `devcontainer-feature.json` schema, which is very close to a `devcontainer.json` but distinct.  The goal of the `$

avidal:
Another question, since this is still in review: this proposal adds an `$origin` for lifecycle commands when they are merged, as well as `$entrypoints` to collect entrypoints from features. What about other properties that are merged, such as `containerEnv`, `forwardPorts`, or `mounts`? What about last-value such as `containerUser` or the specific semantics of `init`?

Do we think it's just not important to be able to report on the origin of those properties? Whereas lifecycle commands can use the origin for output logging?

proposals/consolidate-merged-lifecycle-hook-schema.md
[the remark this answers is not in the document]
avidal: > The goal of the `$` properties as members of the `devcontainer.json` is so that a command such as `devcontainer read-configuration` can output a json object that _is_ also a devcontainer.json. That's not the case today without these changes.

I've been thinking more about this, and I'm not sure that's a good thing to have in the spec itself...it seems like an implementation detail for the specific tool in question. The way some of the other comments read indicates that it'd be "valid" in a `devcontainer.json` but not actually used (re: a comment above about `$entrypoints`)...which kind of implies that the output of `read-configuration` is not *actually* a `devcontainer.json`.

That is, it seems like a "proper" implementation would use this schema to encode/decode metadata stored in the image label, and use it internally as the result of merging all of the various metadata sources, but it's not actually valid `devcontainer.json` config. Effectively splitting out merged metadata into a separate schema and leaving the existing `devcontainer.json` and `devcontainer-feature.json` alone.

The way I've been thinking about it lately is that a `*.json` contains two sets of data: config and metadata. I parse either `devcontainer.json` or `devcontainer-feature.json` into `Config` or `FeatureConfig` respectively, plus a `Metadata` entry. I parse the image label into a `MergedMetadata` entry. Internally I store the config and the merged metadata to drive the build process. I use the merged metadata to render into an image label. I print the config and merged metadata during debug output as two separate entities.

Skaillet91 (changes requested):
[]()

Issue #109, opened 2022-10-02
features that depend upon features (composite features)


johnnyreilly:
Heya!

I'm a keen user of dapr.  I generally do this in the context of a devcontainer.  I'm interested in how to migrate to doing the same using features.  If I read things right, I need to use the `docker-in-docker` feature.  On top of that I think I'd need to have some kind of `dapr` feature which would contain something like this:

```
RUN apt-get update \
    && apt-get install python3-pip -y \
    # Install Dapr
    && wget -q https://raw.githubusercontent.com/dapr/cli/master/install/install.sh -O - | /bin/bash \
    # Clean up
    && apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/* 

# Add daprd to the path for the VS Code Dapr extension.
ENV PATH="${PATH}:/home/${USERNAME}/.dapr/bin"
```

I'm guessing that there's some way to have "dependencies" for features?  So the dapr feature would depend upon docker-in-docker.  Am I headed in the right direction here?  Or am I ... not?

It's possible that this is related: https://github.com/devcontainers/cli/issues/163

AlexanderLanin:
I was missing hard-dependencies since I started experimenting with features, but actually that would be a bad idea. The whole point of features is to avoid god like Dockerfiles. We want small reusable pieces. Dependencies between features make it impossible to use a different feature to install python. Or maybe it's already in my base image and I don't even need a feature at all.

My solution for the moment is to detect whether python is installed and in the correct version. If not, my feature exits with an error message telling the user to add python to the Dockerfile or to use \*any\* python feature.

Assuming features are named reasonably, what I would love is to have `installsAfter: "*python*"`

joshspicer:
> Assuming features are named reasonably, what I would love is to have installsAfter: "*python*"

Using a regex is an interesting idea.  It would be assuming a lot, but I bet it would be right the majority of the time (and I suppose _any_ hint is better than random?)

cc/ @chrmarti  If not regexes, could be interesting to explore how else to drop general hints about install order.

AlexanderLanin:
How about fallback to dependencies? Maybe that's some middle ground between hard dependencies and soft dependency order?

What I currently do:
```sh
if ! type python3 >/dev/null 2>&1; then
  echo "ERROR: Use a base image with python3 installed, add it to your Dockerfile or use a python feature like https://github.com/devcontainers/features/tree/main/src/python"
  exit 1
fi
```

[...]

PR #208, opened 2023-03-08
[proposal] Composite Features

joshspicer:
ref: https://github.com/devcontainers/spec/issues/109

A proposal for a Composite Dev Container Feature.

### DIFF
--- a/proposals/features-composite-dependencies.md
+++ b/proposals/features-composite-dependencies.md
@@ -46,7 +46,7 @@ Introduce a new file type `devcontainer-feature.composite.json` with the followi
 | `id` | `string` | The ID of the Feature.  This follows the same semantics of the `id` property in the `devcontainer-feature.json` file. |
 | `version` | `string` | The version of the Feature.  This follows the same semantics of the `version` property in the `devcontainer-feature.json` file. |
 | `features` | `array` | An array of objects (in installation order) that define the Feature(s) that compose this Feature. |
-| `features.id` | `string` | The ID of the Feature that this Feature depends on. |
+| `features.id` | `string` | The ID of the Feature that this Feature depends on.  A Feature here can be from the same, or different namespaces/repos. |
 | `features.version` | `string` | The version of the Feature that this Feature depends on. |
 | `features.detect` | `string` | A command that will be executed in a shell to determine if the Feature should be installed.  If the command returns a non-zero exit code, the Feature will be installed.  If the command returns a zero exit code, the remaining install steps will be skipped. |
 | `features.options` | `object` | An object of key/value pairs that will be passed to the Feature's `install.sh` script. |


proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Specification
+
+Introduce a new file type `devcontainer-feature.composite.json` with the following properties.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `id` | `string` | The ID of the Feature.  This follows the same semantics of the `id` property in the `devcontainer-feature.json` file. |
+| `version` | `string` | The version of the Feature.  This follows the same semantics of the `version` property in the `devcontainer-feature.json` file. |
+| `features` | `array` | An array of objects (in installation order) that define the Feature(s) that compose this Feature. |
```
chrmarti, line 46: Does the correct installation order depend on each of the install script's requirements?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+| `features.id` | `string` | The ID of the Feature that this Feature depends on. |
```
chrmarti, line 47: Can these be from a different repository/namespace?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+| `features.version` | `string` | The version of the Feature that this Feature depends on. |
```
chrmarti, line 48: What if a different version is configured in the devcontainer.json?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+| `features.detect` | `string` | A command that will be executed in a shell to determine if the Feature should be installed.  If the command returns a non-zero exit code, the Feature will be installed.  If the command returns a zero exit code, the remaining install steps will be skipped. |
```
chrmarti, line 49: Does "remaining install steps will be skipped" mean, the remaining features listed in the array will be skipped?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+| `features.options` | `object` | An object of key/value pairs that will be passed to the Feature's `install.sh` script. |
```
chrmarti, line 50: How are these merged when the same feature is installed by the devcontainer.json?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+            "detect": undefined, // Omit or set as 'undefined' to always install this Feature.
+            "options": {
+                "zip": "zap"
+            }
+        }
+    ]
+}
+```
+
+An optional `finalize.sh` script can be included, and will be fired after all Features have been installed.
```
chrmarti, line 79: Do we have good examples for that?

chrmarti (changes requested):

proposals/features-composite-dependencies.md
joshspicer: Yes, I see no reason to restrict if this proposal is followed.

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
+# Composite Features
+
+Related: https://github.com/devcontainers/spec/issues/109
+
+## Motivation
+
+We've seen significant interest in the ability to "reuse" or "extend" a given Feature with one or more additional Features.  Often in software a given tool depends on another (or several) tool(s)/framework(s).  As the dev container Feature ecosystem has grown, there has been a growing need to reduce redundant code in Features by utilizing first installing a set of dependant Features.
```
bamurtaugh, line 7:
```suggestion
We've seen significant interest in the ability to "reuse" or "extend" a given Feature with one or more additional Features.  Often in software a given tool depends on another (or several) tool(s)/framework(s).  As the dev container Feature ecosystem has grown, there has been a growing need to reduce redundant code in Features by first installing a set of dependent Features.
```

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+
```
bamurtaugh, line 8: Would we want to link to https://github.com/devcontainers/spec/pull/209 as well, since both are related to sharing or reusing Feature code in some way?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+## Goal
+
+The solution shall provide a way to publish a Feature that "depends" on >= 1 other pubished Features. Dependent Features will be installed by the ochestrating tool, with the version and order set by the author if necessary.
```
bamurtaugh, line 11:
```suggestion
The solution shall provide a way to publish a Feature that "depends" on >= 1 other published Features. Dependent Features will be installed by the orchestrating tool, with the version and order set by the author if necessary.
```

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+                "zip": "zap"
+            }
+        }
+    ]
+}
+```
+
+An optional `finalize.sh` script can be included, and will be fired after all Features have been installed.
+
+A composite Feature will be published following the same process as an [standalone dev container Feature](https://containers.dev/implementors/features) into the same namespace - following the pattern outlined in [the Features distribution spec](https://containers.dev/implementors/features-distribution/). Dependencies of a composite Feature can be published to the same or different namespaces.
```
bamurtaugh, line 81:
```suggestion
A composite Feature will be published following the same process as a [standalone dev container Feature](https://containers.dev/implementors/features) into the same namespace - following the pattern outlined in [the Features distribution spec](https://containers.dev/implementors/features-distribution/). Dependencies of a composite Feature can be published to the same or different namespaces.
```

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+
+An example repository structure for a repo with one composite Feature and an standalone Feature can be found below:
```
bamurtaugh, line 83:
```suggestion
An example repository structure for a repo with one composite Feature and a standalone Feature can be found below:
```

proposals/features-composite-dependencies.md
bamurtaugh: This section reminds me that we may want to update the Feature distribution docs with some note that distributing composite Features follows the same process as standalone Features.

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+│   ├── a
+│   │   ├── README.md
+│   │   ├── devcontainer-feature.json
+│   │   └── install.sh
+...
+```
+
+### Notes:
+
+- Composite Features cannot depend on other composite Features. This is to prevent a circular dependencies and deep dependency chains.
```
bamurtaugh, line 102:
```suggestion
- Composite Features cannot depend on other composite Features. This is to prevent circular dependencies and deep dependency chains.
```

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+- The `dependsOn` array is in the order that the Features should be installed.  This is important for Features that depend on the existence of a file or directory created by a previous Feature.
```
bamurtaugh, line 103: Where does `dependsOn` come from (as this is the first time it's mentioned in this doc)? Is it up to the user to determine the order?

proposals/features-composite-dependencies.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Advantages
+
+- Composite Features can be used to distribute a single Feature that depends on other Features.
+- Composite Features can pin all of their dependencies to a specific version, ensuring that the Feature can be tested and will work as expected.
+- Composite Features prevent the complexity that arises with deeply nested dependencies or circular dependencies.
+
+## Disadvantages
+
+- Composite features are not as flexible as other possible dependency models.
```
bamurtaugh, line 117: Do you have an example in mind of how other models are more flexible, so that readers can more fully understand which limitation(s) this refers to?

bamurtaugh (commented):
Very thorough and clear, thanks for putting this together! Left a couple minor comments and questions.

proposals/features-composite-dependencies.md
joshspicer: This was from an earlier draft. Removing for now.

proposals/features-composite-dependencies.md
joshspicer: This could be omitted in the first pass if we find it to be unnecessary or without justification yet.  I'll leave for discussion

proposals/features-composite-dependencies.md
joshspicer: I will clarify this - I meant that _each Feature_ will have a `detect` that will cause the further installation and merging of dev container config **for that Feature**.  Remaining Features in the array will still be executed (and their detect phase run, too).

proposals/features-composite-dependencies.md
joshspicer: We'd want to peek at each Feature's `installsAfter`, i think. That was the primary reason that I didn't want to (for now) enable nesting composite Features.   This spec does NOT let you nest one composite feature in another, making this process easier to resolve (we can "flatten" the installation from composite Features into one install order.

This is something to discuss and finalize in the spec, but we'd probably follow `installsAfter` installation order with precedence within a composite Feature first, and then some algorithm for ordering composite Features based on that?  Perhaps the other way around?

joshspicer:
📢 Open Questions

- [ ]  [Include a `finalize.sh` script?](https://github.com/devcontainers/spec/pull/208#discussion_r1135618671)

- [ ] [Behavior when installing the same Feature >1 times? (eg: embedded in composite Feature and in user dev container)](https://github.com/devcontainers/spec/pull/208#discussion_r1149806033)

  - [ ] [_(Related to above)_ Installing the same Feature(s), different versions?](https://github.com/devcontainers/spec/pull/208#discussion_r1140474890)
 
- [ ] [Install order precedence with existing mechanisms (eg: `installsAfter`)](https://github.com/devcontainers/spec/pull/208#discussion_r1135612733)

proposals/features-composite-dependencies.md
joshspicer: Similar to https://github.com/devcontainers/spec/pull/208#discussion_r1140474890

proposals/features-composite-dependencies.md
chrmarti: Since we are already thinking about composite features depending on each other: Wouldn't it be simpler to avoid the additional artifact type of a "composite feature" and instead allow features to have hard dependencies in addition to the soft dependencies supported with "installAfter"? We already compute the dependency graph including cycle detection for "installAfter". (Other package management systems also seem to have only one package type that can carry code/binaries and dependencies. Do we have an example that keeps these apart?)

We could add "dependencies" to the feature metadata as an object like "features" in the devcontainer.json, so installation order is decided by the same dependency graph algorithm. A dependency would implicitly go into the "installAfter" list of the feature depending on it.

proposals/features-composite-dependencies.md
chrmarti: What are the examples we want to cover with this? Would this run after all previous features have detected & installed, immediately before this feature's install script would run?

proposals/features-composite-dependencies.md
chrmarti: I like that we are discussing various options. We should aim for as much simplicity as we can while still covering the important cases we have. It will be easier to later add to a simple design than a complex one.

nanis94: main

nanis94:
main

joshspicer:
closing in favor of https://github.com/devcontainers/spec/pull/234

Issue #129, opened 2022-10-16
Share code between features


AlexanderLanin:
# Motivation
While devcontainers features aim to enable modern software development, the feature code itself is stuck in a lot of code duplication.
For example we can find `# Determine the appropriate non-root user` 17 times within the features repo. It's the same for me for detecting python version etc etc.

One of the main principles in software development is DRY: Don't Repeat Yourself. That is currently not possible for features!

# Proposal
## Variant: generic pre-build
Add support for pre-build.sh (in src/_pre_build.sh, src/_pre_build/pre_build.sh, src/_special_cli_stuff/pre_build.sh) which would be executed upon build and feature test calls with an appropriate set of parameters. This way the script can do arbitrary things like copy common code. But it could do even more like downloading common code from yet another source. It could merge code into single files. etc.

## Variant: common code
Add support for src/_common which could contain code that is used throughout my features. Build and feature test calls will package that code as well as the install.sh script of the feature. Within install.sh one can simply source/call anything from _common as it's part of the package.

*P.S. I wasn't sure whether to put this to cli or spec, please move if incorrect.*

AlexanderLanin:
19 days have passed and I have a new view on that. That's how fast the software world changes ;-)

Current setup unintentionally allows a powerful reuse of those scripts we have, for example:
`curl -s https://raw.githubusercontent.com/devcontainers/features/main/src/python/install.sh | sudo VERSION=3.7 bash`

joshspicer:
+1 to this.  Below is just some spitballing, but an idea i've been tossing around is the concept of _library Features_.  They cannot be referenced directly in a devcontainer.json, but rather a Feature will declare that it depends on a _library_.


```json
{
   "id": "ruby":,
   "version": 1.2.3", 
    "libraries": [
      "ghcr.io/devcontainers/features/lib:1"
    ]
}
```

The CLI would then automatically fetch `ghcr.io/devcontainers/features/lib:1` and source the contents into the executing `install.sh` script.  These libs would be cached as any other OCI artifact, but allows an author to quickly update "helper functions" for their entire suite of Features.

For the devcontainers maintainers team, we would publish and maintain a `ghcr.io/devcontainers/features/lib` _library Feature_ that anyone could use.  For our own Features in `devcontainers/features`, it would reduce the size of each `install.sh` _significantly_ and reduce the copy/paste errors and out-of-sync errors we occasionally see.

jcbhmr:
[...]

PR #209, opened 2023-03-09
[proposal] Sharing code between Dev Container Features

joshspicer:
ref: https://github.com/devcontainers/spec/issues/129

### DIFF
--- a/proposals/features-library.md
+++ b/proposals/features-library.md
@@ -14,25 +14,53 @@ Provide a generic and efficient way to share code between Features.
 
 # Proposal
 
-Add an `include` property to the `devcontainer-feature.json`.  This will be honored during the Feature [packaging stage](https://containers.dev/implementors/features-distribution/#packaging). The contents of the provided relative path is copied into the packaged Feature archive.
+Add an `include` property to the `devcontainer-feature.json`.  This will be honored during the Feature [packaging stage](https://containers.dev/implementors/features-distribution/#packaging). 
+
+The contents of the provided relative path is copied into the packaged Feature archive.  The packaging stage already has a concept of a "root" directory (likely the git root of the directory, containing a `src` and `test` folder), so the `include` property can be relative to that directory.  The relative path must not escape the root directory - if this is attempted, the packaging stage will fail.
 
 Property | Type | Description
 --- | --- | ---
-`include` | `string[]` | An array of relative paths to package with the Feature. Paths are relative to root directory when packaging (often the git root directory).  If the element is a folder, it is copied recursively.  Must be prefixed with `.` or `..` to resolve the relative path.
+`include` | `string[]` | An array of relative paths to package with the Feature. Paths are relative to root directory when packaging (often the git root directory).  If the element is a folder, it is copied recursively.  Must be prefixed with `.` to indicate the provided string is relative path.
 
 ## Example
 
 ```json
 {
-    "name": "my-feature",
+    "name": "featureA",
     "version": "0.0.1",
     "include": [
         "./utils/",
+        "./company-welcome-message.txt"
     ]
 }
 ```
 
-The preceding example will recursively copy the contents of the `utils` folder into the temporary folder used to package the Feature.  The `utils` folder will be included with the published Features.
+The preceding example will recursively copy the contents of the `utils` folder into the temporary folder used to package the Feature.  Additionally, the `company-welcome-message.txt` file will also be packaged with the Feature.  The `utils` folder will be included with the published Features.
+
+The directory stucture may look like the following:
+
+```
+.
+├── company-welcome-message.txt
+├── utils
+|   ├── common.sh
+│   └── helpers
+│       ├── a.sh
+│       └── ...
+|
+├── src
+│   ├── featureA
+│   │   ├── devcontainer-feature.json
+│   │   ├── install.sh
+│   │   └── ...
+│   ├── featureB
+│   │   ├── devcontainer-feature.json
+│   │   └── install.sh
+|   ├── ...
+│   │   ├── devcontainer-feature.json
+│   │   └── install.sh
+├── ...
+```
 
 This functionality will enable sharing a common set of helper functions between Features.  The following example shows how a Feature could include a `utils` folder that contains a `common.sh` file.
 
@@ -44,7 +72,7 @@ function common_function() {
 ```
 
 ```bash
-# my-feature/install.sh
+# featureA/install.sh
 
 # Include common functions
 source "./utils/common.sh"
@@ -68,7 +96,7 @@ A "library" can be included in a Feature's `include` property as an alternative 
 ```json
 
 {
-    "name": "my-feature",
+    "name": "featureA",
     "version": "0.0.1",
     "include": [
         "ghcr.io/devcontainers/features/utils:0.0.1"
--- a/proposals/features-library.md
+++ b/proposals/features-library.md
@@ -10,17 +10,21 @@ Additionally, [members of the Feature authoring community have expressed interes
 
 # Goal
 
-Provide a generic and efficient way to share code between Features.
+Provide a generic and simple way to share code between Features.
 
-# Proposal
+# Proposals
+
+## [Proposal A] Add: `include` property to `devcontainer-feature.json`
 
 Add an `include` property to the `devcontainer-feature.json`.  This will be honored during the Feature [packaging stage](https://containers.dev/implementors/features-distribution/#packaging). 
 
-The contents of the provided relative path is copied into the packaged Feature archive.  The packaging stage already has a concept of a "root" directory (likely the git root of the directory, containing a `src` and `test` folder), so the `include` property can be relative to that directory.  The relative path must not escape the root directory - if this is attempted, the packaging stage will fail.
+The contents of the provided relative path is copied into the packaged Feature archive.  Files or directories are recursively copied into the archive following the same directory structure as in the source code.
+
+The relative path must not escape the root directory - if this is attempted, the packaging stage will fail.
 
 Property | Type | Description
 --- | --- | ---
-`include` | `string[]` | An array of relative paths to package with the Feature. Paths are relative to root directory when packaging (often the git root directory).  If the element is a folder, it is copied recursively.  Must be prefixed with `.` to indicate the provided string is relative path.
+`include` | `string[]` | An array of paths relative to the directory containing this `devcontainer-feature.json`. If the element is a folder, it is copied recursively.  Must be prefixed with `.` to indicate the provided string is relative path.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


proposals/features-library.md
```diff
@@ -0,0 +1,76 @@
...
+
+Provide a generic and efficient way to share code between Features.
+
+# Proposal
+
+Add an `include` property to the `devcontainer-feature.json`.  This will be honored during the Feature [packaging stage](https://containers.dev/implementors/features-distribution/#packaging). The contents of the provided relative path is copied into the packaged Feature archive.
+
+Property | Type | Description
+--- | --- | ---
+`include` | `string[]` | An array of relative paths to package with the Feature. Paths are relative to root directory when packaging (often the git root directory).  If the element is a folder, it is copied recursively.
```
jkeech, line 21: What would happen for paths that resolve outside of the feature root directory, such as `../shared/utils/`?

proposals/features-library.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Possible Extensions
+
+This section provides hints to possible future work that could extend the proposal above.  These are not part of the proposal, but are provided to help the community understand the potential future impact of this proposal.
+
+## Distributing shared library Features
+
+In an effort to reduce the need for authors to copy/paste auxillary scripts between Feature namespaces, provide the ability to publish "library" Features.  These Features would be published [following the Feature distribution spec](https://containers.dev/implementors/features-distribution/#distribution) under a new manifest type. 
+
+A "library" can be included in a Feature's `include` property as an alterative to a relative path. By themselves, a "library" Feature would not be installable.
```
jkeech, line 66: Similar, for this library feature, how would consumers locate these scripts on disk? Or would they be automatically sourced before execution of the consuming feature script?

proposals/features-library.md
```diff
@@ -0,0 +1,76 @@
...
+function common_function() {
+   echo "Hello, '$1'"
+}
+```
+
+```bash
+# my-feature/install.sh
+
+# Include common functions
+source "./utils/common.sh"
```
chrmarti, line 50: I guess this is `./utils` and not `../utils` because the packaged folder structure is different from the folder structure in the repository? So this can't be run from within the repository because the paths do not match?

Wouldn't it make development of features easier if the folder structures were the same? E.g., can we make it such that features can be run from the repository as "local" features during feature development?

@samruddhikhandale Any insights from your work on features?

chrmarti (changes requested):

proposals/features-library.md
joshspicer: Similar to how we don't resolve Local Features outside of the `.devcontainer` folder, I propose failing the packaging step.

proposals/features-library.md
joshspicer: Yes, that's how I was originally imagining it.  Anything in the `includes` would be copied recursively to the root of the Feature's packaged directory.

proposals/features-library.md
joshspicer: I think that's open for debate whether the user has the choice to source, or whether we try to auto-source whatever is included (might be trickier if we copy recursively, so perhaps we'd need to limit to include explicit script files)

joshspicer: My example below has the consumer source explicitly at the top of their `install.sh`.

proposals/features-library.md
```diff
@@ -0,0 +1,76 @@
+# Sharing code between Dev Container Features
+
+ref: https://github.com/devcontainers/spec/issues/129
+
+# Motivation
+
+The current model for authoring Features encourages significant copy and paste between individual Features.  Most of the Features published under the `devcontainers/features` namespace contain approximately 100 lines of [identifical helper functions](https://github.com/devcontainers/features/blob/main/src/azure-cli/install.sh#L29-L164) before the first line of the script is executed.  The process of maintaining these helper functions is tedious, error-prone, and leads to less readable Features.
```
bamurtaugh, line 7:
```suggestion
The current model for authoring Features encourages significant copy and paste between individual Features.  Most of the Features published under the `devcontainers/features` namespace contain approximately 100 lines of [identical helper functions](https://github.com/devcontainers/features/blob/main/src/azure-cli/install.sh#L29-L164) before the first line of the script is executed.  The process of maintaining these helper functions is tedious, error-prone, and leads to less readable Features.
```

proposals/features-library.md
```diff
@@ -0,0 +1,76 @@
...
+{
+    "name": "my-feature",
+    "version": "0.0.1",
+    "include": [
+        "./utils/",
+    ]
+}
+```
+
+The preceeding example will recursively copy the contents of the the `utils` folder into the temporary folder used to package the Feature.  The `utils` folder will be included with the published Features.
```
bamurtaugh, line 35:
```suggestion
The preceding example will recursively copy the contents of the `utils` folder into the temporary folder used to package the Feature.  The `utils` folder will be included with the published Features.
```

proposals/features-library.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Possible Extensions
+
+This section provides hints to possible future work that could extend the proposal above.  These are not part of the proposal, but are provided to help the community understand the potential future impact of this proposal.
+
+## Distributing shared library Features
+
+In an effort to reduce the need for authors to copy/paste auxillary scripts between Feature namespaces, provide the ability to publish "library" Features.  These Features would be published [following the Feature distribution spec](https://containers.dev/implementors/features-distribution/#distribution) under a new manifest type. 
+
+A "library" can be included in a Feature's `include` property as an alterative to a relative path. By themselves, a "library" Feature would not be installable.
```
bamurtaugh, line 66:
```suggestion
A "library" can be included in a Feature's `include` property as an alternative to a relative path. By itself, a "library" Feature would not be installable.
```

bamurtaugh (commented):
Thanks so much for putting this together! Very clear, and I think it's great to address this due to the motivations and feedback you mention. Only had very nit cleanup comments.

proposals/features-library.md
chrmarti: I would suggest to keep the folder structure the same as in the repository. The main difficulty from what I see is the lookup of the `devcontainer-feature.json` which is currently expected at the root of the archive. If we know the feature id, we could additionally check in the folder with the same name as the unqualified id.

proposals/features-library.md
joshspicer: Added more detail

proposals/features-library.md
joshspicer: For the Feature development case I don't think that makes too much of a difference - the `features test` command currently does a full copy into a temp folder inside of a temporary project's `.devcontainer` (to make it a 'local Features' from the perspective of the CLI).  This could likely be optimized, but we have strict rules over where local Features can live relative to the devcontainer.json, so it's a bit of hassle to _not_ do the copying.  I see the appeal of keeping the directory structure, and depending on if a 'local' Feature or a 'packaged' Feature, change where we look for the `include` files.  That said, we haven't really gotten feedback that performance on the testing command is an issue, and the vast majority of the time is not spent setting up the tests.

I don't know how useful that behavior is in practice, and makes figuring out the correct path slightly more confusing when trying to use an included script in a packaged Feature.

Also, how would it work sourcing from `../utils`.  Would we need to re-adjust how the entire archive is structured (ie: create a dummy `src` folder that has one folder `feature-name`, that contains `install.sh` and `devcontainer-feature.json`?

proposals/features-library.md
joshspicer: I see that there are two options:

- No implicit sourcing.  The `include` simply copies the files into the archive.  It's up to the author to decide how and where to source/execute/copy those resources.
-  Treat script files (likely by looking for `.sh` and potentially other extensions) differently, and automatically source them in the feature-install.sh preamble script before executing the author's `install.sh`

joshspicer:
📣 Open Questions

- [ ]  [Auto source scripts into `install.sh` or manually source?](https://github.com/devcontainers/spec/pull/209#discussion_r1139430977)
- [ ]  [What is the path is copied scripts relative to?](https://github.com/devcontainers/spec/pull/209#discussion_r1143595780)

proposals/features-library.md
chrmarti: No implicit sourcing is more flexible (e.g., scripts might be written like commands that immediately do something). 👍

proposals/features-library.md
```diff
@@ -72,15 +101,55 @@ function common_function() {
...
+
+## [Proposal B] Change: Structure of packaged Feature
+
+To simplify the proposed change above and keep the relative paths consistent before and after packaging, the structure of the packaged Feature will change to mirror the directory structure of the source code.
+
+When running the packaging step, the stage is provided a source directory for packaging >1 Feature, or the root of a project when packaging a single Feature.  The Feature contents will be placed into `./src/<FeatureID>`.  Other included directories will be copied into the root of the packaged artifact as indicated in **Proposal A**.
+
+An implementation should resolve the `devcontainer-feature.json` by checking for its presence in the following order of precedence:
+
+ - `./src/<feature-id>/devcontainer-feature.json`
```
chrmarti, line 122: I wonder if we need to set `src` in stone. We could copy the `devcontainer-feature.json` to the root folder - probably updating relative paths in its properties, so they do not break. Or: Is it possible to have a symbolic link inside the tar file, making `devcontainer-features.json` at the root a link to the actual location?

chrmarti, line 122: I wonder if we need to set `src` in stone. We could copy the `devcontainer-feature.json` to the root folder - probably updating relative paths in its properties, so they do not break. Or: Is it possible to have a symbolic link inside the tar file, making `devcontainer-features.json` at the root a link to the actual location?

chrmarti (changes requested):

proposals/features-library.md
```diff
@@ -72,15 +101,55 @@ function common_function() {
...
+source "../../utils/common.sh"
 
 # Use common function
 common_function "devcontainers"
 ```
 
+
+## [Proposal B] Change: Structure of packaged Feature
+
+To simplify the proposed change above and keep the relative paths consistent before and after packaging, the structure of the packaged Feature will change to mirror the directory structure of the source code.
```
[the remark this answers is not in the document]
joshspicer: Thanks, I removed some duplication.

proposals/features-library.md
joshspicer: > I wonder if we need to set src in stone.

When publishing you're able to specify any folder as the 'src', ie:

`devcontainers features publish ~/josh/repos/features/src`

Is how i'd publish Features in a repo `features` with a folder `src` on my local machine.

_That_ repo could have any folder name instead of `src`.  I think, for simplicity in looking up the `devconatiner-feature.json`, we'd want to remap whatever that folder is to `src` (as I outlined above), but the user could keep that folder named to whatever they'd like it do be.

Not sure if that resolves your concern at all, just wanted to perhaps provide more context that the _user_ wouldn't necessarily need a folder named `src`, but internally we'd copy Feature's root folder (`<some-user-src-folder>/featureA/...` -> src/featureA...`

joshspicer: > Is it possible to have a symbolic link inside the tar file, making devcontainer-features.json at the root a link to the actual location?

IMO that sounds like added complication.  Not sure, I'd need to do some digging to see how `tar` would react to this on different platforms.

chrmarti (approved):

Issue #198, opened 2023-02-07
Declarative secrets


bdmac:
### Proposal
Allow secrets to be declared in devcontainer outlining requirements to successfully run the container. All declared secrets in this initial specification would be injected into the container as ENV variables.

Declared secrets would be _recommendations_ not _requirements_. Missing secrets should not prevent creation of the container.

### Example

```
{ 
  "secrets": [
    { "name": "MY_SECRET", "description": "This is a great description of MY_SECRET." }
  ]
}
```

Related to https://github.com/github/codespaces/issues/12040

joshspicer:
Would an optional boolean `required` for each secret make sense to include (defaults to `false`) ?

lostintangent:
If a secret represents an external API key, then it might be helpful to allow the repo maintainer to include the URL to retrieve it. That way, the experience can feel as immediately actionable as possible (e.g. "You need this secret and here's where to get it"). 

I could imagine that being enable by...

1. Allowing the `description` to be Markdown (and we render links in the UI)
2. Adding an optional `url` property to the secret schema.

chrmarti:
I assume this is for the devcontainer.json? Would the devcontainer-feature.json also support the declaration of secrets? Would they go into the image metadata?

Not every UI might be able to or allow for rendering Markdown / HTML (e.g., VS Code's QuickPick UI). We could keep the description plain-text and add an optional `"documentationURL"` property to allow for more flexibility in the UI.

I suggest we use a JSON object mapping secret names to their properties, that will align well with existing properties like `"containerEnv"` and `"remoteEnv"`:
```json
    "secrets": {
        "MY_SECRET": {
            "description": "This is a great description of MY_SECRET."
        }
    },
```

Instead of making secrets implicitly part of the environment variables, we could use a similar syntax to the one we are using, e.g., to include local variables `${localEnv:SOME_VARIABLE}` in the devcontainer.json and reference secrets as `${secret:MY_SECRET}`. That would allow for additional flexibility in where and how secrets can be used:
```json
{
    "image": "mcr.microsoft.com/devcontainers/base:bullseye",
    "secrets": {
        "MY_SECRET": {
            "description": "This is a great description of MY_SECRET."
        }
    },
    "remoteEnv": {
        "MY_SECRET": "${secret:MY_SECRET}",
        "SOME_VARIABLE": "${localEnv:SOME_VARIABLE}"
    },
    "features": {
        "ghcr.io/devcontainers/features/my_feature:1": {
[...]

PR #214, opened 2023-03-16
Add a proposal for declarative secrets

bdmac:
ref: https://github.com/devcontainers/spec/issues/198

### DIFF
--- a/proposals/declarative-secrets.md
+++ b/proposals/declarative-secrets.md
@@ -8,19 +8,19 @@ Various projects exist in the wild that require various secrets for them to run 
 - https://github.com/openai/openai-quickstart-python
 - https://github.com/openai/openai-quickstart-node
 
-Today these projects have to include specific instructions e.g. in their README telling users where to procure what they need to run and then how to set it up as a secret or add it as an environment variable. This currently acts as an impediment to adoption and promotion of codespaces for these projects.
+Today these projects have to include specific instructions e.g. in their README telling users where to procure what they need to run and then how to set it up as a secret or add it as an environment variable. This currently acts as an impediment to adoption and promotion of dev containers for these projects.
 
 ## Goal
 
-Simplify using codespaces for these kinds of projects by including it as a first-class part of the codespace creation flow.
+Simplify using dev containers for these kinds of projects by supporting secrets as a first-class part of the dev container creation flow.
 
 ## Proposal
 
-Add an optional `secrets` to the `devContainer.base.schema.json`. This will be used to declare the secrets needed within the devcontainer.
+Add an optional `secrets` to the `devContainer.base.schema.json`. This will be used to declare the secrets needed within the dev container.
 
 Property | Type | Description
 --- | --- | ---
-`secrets` | `object` | The keys of this object are the names of the secrets that are in use.
+`secrets` | `object` | The keys of this object are the names of the secrets that are in use. Keys should be [valid Linux environment variable names](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html).
 
 The `secrets` property contains a map of secret names and details about those secrets. These `secrets` are recommended to the user but are not _required_ for creation. All properties are optional.
 
@@ -44,15 +44,13 @@ Property | Type | Description
 }
 ```
 
-This example would declare that this devcontainer wants the user to provide two secrets. The `CHAT_GPT_API_KEY` secret also provides some additional metadata that clients can use in displaying a user experience to provide that secret. The `STABLE_DIFFUSION_API_KEY` skips that metadata.
+This example would declare that this dev container wants the user to provide two secrets. The `CHAT_GPT_API_KEY` secret also provides some additional metadata that clients can use in displaying a user experience to provide that secret. The `STABLE_DIFFUSION_API_KEY` skips that metadata.
 
-These secrets would, by default, be injected into the container as environment variables.
-
-For codespaces specifically we will prompt the user for this information as part of creating their codespace. The `description` and `documentationUrl` are leveraged in the UI to provide context to the user about the secret. As mentioned above the secrets would be _optional_ as part of creating a codespace.
+Implementations _may_ inject these secrets into the container as environment variables.
 
 ## Possible Future Extensions
 
 These are **out of scope** for this proposal but the following has been discussed in relation:
 
-1. Inclusion of a `type` field per secret. The default, which we are assuming and leveraging here, would be `"environmentVariable"`. This would include an option to specify `"reference"` and then reference the provided secrets in other parts of the devcontainer.
-2. Prescriptions of _how_ the secrets are injected into the devcontainer.
+1. Inclusion of a `type` field per secret. The default, which we are assuming and leveraging here, would be `"environmentVariable"`. This would include an option to specify `"reference"` and then reference the provided secrets in other parts of the dev container.
+2. Prescriptions of _how_ the secrets are injected into the dev container.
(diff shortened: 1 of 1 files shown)


proposals/declarative-secrets.md
```diff
@@ -0,0 +1,58 @@
...
+
+Simplify using codespaces for these kinds of projects by including it as a first-class part of the codespace creation flow.
+
+## Proposal
+
+Add an optional `secrets` to the `devContainer.base.schema.json`. This will be used to declare the secrets needed within the devcontainer.
+
+Property | Type | Description
+--- | --- | ---
+`secrets` | `object` | The keys of this object are the names of the secrets that are in use.
```
joshspicer, line 23: Should we note here the shape of what a key in this map should look like?

Since implementations of this will likely source these key as environment variables, I think it would make sense to align with what a valid env variable key looks like (in unix-like systems): https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names#2821183

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,58 @@
...
+        "CHAT_GPT_API_KEY": {
+            "description": "I'm your super cool API key for ChatGPT.",
+            "documentationUrl": "https://openai.com/api/"
+        },
+        "STABLE_DIFFUSION_API_KEY": {}
+    }
+}
+```
+
+This example would declare that this devcontainer wants the user to provide two secrets. The `CHAT_GPT_API_KEY` secret also provides some additional metadata that clients can use in displaying a user experience to provide that secret. The `STABLE_DIFFUSION_API_KEY` skips that metadata.
```
joshspicer, line 47:
```suggestion
This example would declare that this dev container wants the user to provide two secrets. The `CHAT_GPT_API_KEY` secret also provides some additional metadata that clients can use in displaying a user experience to provide that secret. The `STABLE_DIFFUSION_API_KEY` skips that metadata.
```

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,58 @@
...
+
+These secrets would, by default, be injected into the container as environment variables.
```
joshspicer, line 49: Not sure if this is necessary to state? Could be convinced otherwise but this sounds like an implementation detail.

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,58 @@
...
+
+For codespaces specifically we will prompt the user for this information as part of creating their codespace. The `description` and `documentationUrl` are leveraged in the UI to provide context to the user about the secret. As mentioned above the secrets would be _optional_ as part of creating a codespace.
+
+## Possible Future Extensions
+
+These are **out of scope** for this proposal but the following has been discussed in relation:
+
+1. Inclusion of a `type` field per secret. The default, which we are assuming and leveraging here, would be `"environmentVariable"`. This would include an option to specify `"reference"` and then reference the provided secrets in other parts of the devcontainer.
+2. Prescriptions of _how_ the secrets are injected into the devcontainer.
```
joshspicer, line 58:
```suggestion
2. Prescriptions of _how_ the secrets are injected into the dev container.
```

joshspicer (approved):

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,58 @@
...
+
+Today these projects have to include specific instructions e.g. in their README telling users where to procure what they need to run and then how to set it up as a secret or add it as an environment variable. This currently acts as an impediment to adoption and promotion of codespaces for these projects.
+
+## Goal
+
+Simplify using codespaces for these kinds of projects by including it as a first-class part of the codespace creation flow.
+
+## Proposal
+
+Add an optional `secrets` to the `devContainer.base.schema.json`. This will be used to declare the secrets needed within the devcontainer.
```
bamurtaugh, line 19: nit: We refer to them as 2 words "dev containers," so would be great to update here and any others spots in the doc 😄

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,56 @@
+# Declarative Secrets
+
+## Motiviation
+
+Various projects exist in the wild that require various secrets for them to run properly. Exmaples include:
```
bamurtaugh, line 5:
```suggestion
Various projects exist in the wild that require various secrets for them to run properly. Examples include:
```

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,56 @@
...
+
+Today these projects have to include specific instructions e.g. in their README telling users where to procure what they need to run and then how to set it up as a secret or add it as an environment variable. This currently acts as an impediment to adoption and promotion of dev containers for these projects.
+
+## Goal
+
+Simplify using dev containers for these kinds of projects by supporting secrets as a first-class part of the dev container creation flow.
+
+## Proposal
+
+Add an optional `secrets` to the `devContainer.base.schema.json`. This will be used to declare the secrets needed within the dev container.
```
bamurtaugh, line 19:
```suggestion
Add an optional `secrets` property to the `devContainer.base.schema.json`. This will be used to declare the secrets needed within the dev container.
```

bamurtaugh (commented):
Thanks for putting this together! Looks great, only had a couple nit comments.

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,56 @@
...
+        "CHAT_GPT_API_KEY": {
+            "description": "I'm your super cool API key for ChatGPT.",
+            "documentationUrl": "https://openai.com/api/"
+        },
+        "STABLE_DIFFUSION_API_KEY": {}
+    }
+}
+```
+
+This example would declare that this dev container wants the user to provide two secrets. The `CHAT_GPT_API_KEY` secret also provides some additional metadata that clients can use in displaying a user experience to provide that secret. The `STABLE_DIFFUSION_API_KEY` skips that metadata.
```
chrmarti, line 47: "some additional metadata" means the documentation URL the user can navigate to with the browser or is there anything in addition to that?

proposals/declarative-secrets.md
```diff
@@ -0,0 +1,56 @@
...
+
+Implementations _may_ inject these secrets into the container as environment variables.
```
chrmarti, line 49: To ensure that the same devcontainer.json will result in a functioning dev container with all implementations, shouldn't we _require_ implementations to inject the secrets as environment variables?

chrmarti (changes requested):
Looks good, left a few comments.

proposals/declarative-secrets.md
bdmac: The `description` and `documentationUrl`, yes. It's up to implementations to decide if and how to make use of this metadata.

proposals/declarative-secrets.md
bdmac: I think what this is trying to recognize and call out is that there may be other types of secrets later (not addressed here as mentioned towards the bottom) that would be injected differently and/or not injected at all. This document is implicitly assuming a default value for this TBD future `"type"` property of "environment" but trying to avoid calling that out since I'm not trying to address that possible future inclusion.

FWIW this was originally worded as this but was changed in response to this earlier PR feedback https://github.com/devcontainers/spec/pull/214#discussion_r1139038674
>These secrets would, by default, be injected into the container as environment variables.

chrmarti: "by default" would cover the absence of "type" for me, but we can clarify later if you prefer.

chrmarti (approved):
LGTM, thanks!

PR #222, opened 2023-04-08
Create .devcontainer/devcontainer.json

Wenmeilin:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
```
samruddhikhandale, line 2: Hi 👋 

Can you help me understand the scenario where this dev container is helpful? As the spec repo is used for documentation, I don't think it would need the `universal` image.

joshspicer: Also, the config you've provided is the default image in GitHub Codespaces if you weren't to have a dev container.  

If you've found another case where this change is useful, please re-open this pull request with some info in the description box - thanks!

PR #234, opened 2023-05-02
[Proposal] Feature `dependsOn`

joshspicer:
Proposal for Feature dependencies (if accepted, superseding [Composite Features](https://github.com/devcontainers/spec/pull/208).)

ref: https://github.com/devcontainers/spec/issues/109
reference implementation: https://github.com/devcontainers/cli/pull/530

### DIFF
--- a/proposals/feature-dependencies.md
+++ b/proposals/feature-dependencies.md
@@ -106,6 +106,14 @@ To prevent non-deterministic behavior, the algorithm will sort each round accord
 
 If there is no difference based on these comparator rules, the Features are considered equal.
 
-### (D) Modify existing Feature order installation algorithm
+### NOTE: Existing methods for Feature installation
 
 Two existing properties: `installsAfter` on the Feature metadata, and `overrideFeatureInstallationOrder` in the `devcontainer.json` both exist to alter the installation order of user-defined Features. ::TODO::
+
+### NOTE: Feature authorship
+
+Features should be authored with the following considerations:
+
+- Features should be authored with the assumption that they will be installed in any order, so long as the dependencies are met at some point beforehand.
+- Since two Features with different options are considered different, a single Feature may be installed more than once.  Features should be idempotent.
+- Features that require updating shared state in the container (eg: updating the `$PATH`), should be aware that the same Feature may be run multiple times. Consider a method for previous runs of the Feature to communicate with future runs, updating the shared state in the intended way.
--- a/proposals/feature-dependencies.md
+++ b/proposals/feature-dependencies.md
@@ -156,9 +156,7 @@ If there is no difference based on these comparator rules, the Features are cons
 
 Two existing properties (1) `installsAfter` on the Feature metadata, and  (2) `overrideFeatureInstallationOrder` in the `devcontainer.json` both exist to alter the installation order of user-defined Features.
 
-A Feature may not declare both `installsAfter` and `dependsOn` properties. If a Feature declares both in its metadata, the publishing step should fail with an error.
-
-Furthermore, the dependency resolution outlined in this specification will be first be performed to generate an installation order. If that succeeds, the resolved list of Features will then be processed by the existing properties as previously defined.
+:: TODO :: More discussion needed!
 
 ## Additional Remarks
 
--- a/proposals/feature-dependencies.md
+++ b/proposals/feature-dependencies.md
@@ -1,0 +1,53 @@
+# # Feature Dependencies
+
+Reference: https://github.com/devcontainers/spec/pull/208/files
+
+An alternate and conflicting proposal can be found here: https://github.com/devcontainers/spec/pull/208/.  If one proposal is accepted, the other should be rejected.
+
+## Motivation
+
+We've seen significant interest in the ability to "reuse" or "extend" a given Feature with one or more additional Features.  Often in software a given tool depends on another (or several) tool(s)/framework(s).  As the dev container Feature ecosystem has grown, there has been a growing need to reduce redundant code in Features by first installing a set of dependent Features.
+
+## Goal
+
+The solution shall provide a way to publish a Feature that "depends" on >= 1 other published Features. Dependent Features will be installed by the orchestrating tool following this specification.
+
+The solution outlined shall not only execute the installation scripts, but also merge the additional development container config, as outlined in the documented [merging logic.](https://containers.dev/implementors/spec/#merge-logic).
+
+The solution should provide guidelines for Feature authorship, and prefer solutions that enable simple authoring of Features, as well as easy consumption by end-users.
+
+A non-goal is to require the use or implementation of a full-blown dependency management system (such as `npm` or `apt`).  The solution should allow Features to operate as "self-contained, shareable units of installation code and development container configuration"[(1)](https://containers.dev/implementors/features/). 
+
+## Existing Solutions
+
+See https://github.com/devcontainers/spec/pull/208/files#diff-a29ffaac693437b6fbf001066b97896f7aef4d6d37dc65b8b98b22a5e19f6c7aR26 for existing solutions and alternative approaches.
+
+## Proposed Specification
+
+### (A) Add `dependsOn` property to Feature metadata.
+
+A new property `dependsOn` can be optionally added to the `devcontainer-feature.json`.  This property mirrors the `features` object in `devcontainer.json`.  Adding Feature(s) to this property tell the orchestrating tool to install the Feature(s) before installing the Feature that declares the dependency. 
+
+The installation order is subject to the algorithm set forth in this document. Where there is ambiguity, it is up to the orchestrating tool to decide the order of installation.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `dependsOn` | `string` | The ID of the Feature.  This follows the same semantics of the `id` property in the `devcontainer-feature.json` file. |
+
+An example `devcontainer-feature.json` file with a dependency on three other published Features:
+
+```json
+{
+    "name": "My Feature",
+    "id": "myFeature",
+    "version": "1.0.0",
+    "dependsOn": {
+        "ghcr.io/myotherFeature:1": {
+            "flag": true
+        },
+        "features.azurecr.io/aThirdFeature:1": {},
+        "features.azurecr.io/aFourthFeature:1.2.3": {}
+
+    }
+}
+```
--- a/proposals/feature-dependencies.md
+++ b/proposals/feature-dependencies.md
@@ -142,7 +142,7 @@ Start with all the elements from **(B2)** in a `worklist` and an empty list `ins
 
 Then for each intermediate `round` list, commit to `installationOrder` only those nodes who share the maximum `roundPriority`.  Return all nodes in `round` with a strictly lower `roundPriority` to the `worklist` to be reprocessed in subsequent iterations.  If there are multiple nodes with the same `roundPriority`, commit them to `installationOrder` with a final sort according to **Round Stable Sort**.
 
-Repeat for as many rounds as necessary until the worklist is empty.  If there is ever a round where no elements are added to `installationOrder`, the algorithm should terminate and return an error.  This indicates a circular dependency or other fatal error in the dependency graph.  Implementations should attempt to provide the user with information about the error and possible mitigation strategies.
+Repeat for as many rounds as necessary until `worklist` is empty.  If there is ever a round where no elements are added to `installationOrder`, the algorithm should terminate and return an error.  This indicates a circular dependency or other fatal error in the dependency graph.  Implementations should attempt to provide the user with information about the error and possible mitigation strategies.
 
 
 #### Definition: Feature Equality
--- a/proposals/feature-dependencies.md
+++ b/proposals/feature-dependencies.md
@@ -114,7 +114,7 @@ The orchestrating tool is responsible for calculating a Featuure installation or
 
 #### (B1) Building dependency graph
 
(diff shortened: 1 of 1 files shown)


proposals/feature-dependencies.md
```diff
@@ -0,0 +1,176 @@
...
+  - Sort the provided option values lexicographically. If the comparision is equal:
+- Sort Features by their canonical name (For OCI-published Features, the Feature ID resolved to the digest hash).
+
+If there is no difference based on these comparator rules, the Features are considered equal.
+
+### Comments on existing methods of altering Feature installation order
+
+Two existing properties (1) `installsAfter` on the Feature metadata, and  (2) `overrideFeatureInstallationOrder` in the `devcontainer.json` both exist to alter the installation order of user-defined Features.
+
+:: TODO :: More discussion needed!
```
chrmarti, line 159: Thoughts on using a preprocessed dependency graph that will allow us to run a single ordering algorithm once to cover `dependsOn`, `installsAfter` and `overrideFeatureInstallOrder` of dependencies:
- `installsAfter` and `overrideFeatureInstallOrder` do not include options, so maybe the dependency graph should have nodes for features and two types of directed edges: one for dependencies with options (`dependenciesWithOptions`) and one for dependencies without options (`dependenciesWithoutOptions`). A node's dependencies will be considered 'installed' and the node ready-to-be-installed when all its `dependenciesWithOptions` are installed and for all of its `dependenciesWithoutOptions` at least one with any set of options is installed.
- The `dependsOn` references are added as `dependenciesWithOptions`.
- The `installsAfter` references that are not going to be installed can be omitted from the preprocessed dependency graph. All others are added as `dependenciesWithoutOptions`.
- `overrideFeatureInstallOrder` can be added to the dependency graph with:
  - Add an artificial node for each feature that is in the `overrideFeatureInstallOrder` list with its `dependenciesWithOptions` referring to all nodes of this feature. (This might be optional, but makes the description and maybe the algorithm simpler.)
  - Let each such artificial node depend on the artifical node of the feature preceeding it in the `overrideFeatureInstallOrder` list. (Enforcing the user-provided order.)
  - Let all leave nodes (nodes without any dependencies) in the graph refer to the artificial node of the feature last in the `overrideFeatureInstallOrder` list. (Enforcing that these are installed first.)
- To float user-selected features (those in the devcontainer.json) to the end of the install order, we could assign each node a priority and in each round of the algorithm only handle nodes of the highest (lowest value) available priority. User-selected features would have priority 2, all others priority 1.
  - To make local features float to the end (as in the existing spec), we could assign them priority 3, so they are installed after user-selected features.

chrmarti (commented):
Looks great, added some thoughts.

proposals/feature-dependencies.md
```diff
@@ -0,0 +1,176 @@
...
+}
+```
+
+`myfeature` will be installed after `myotherFeature`, `aThirdFeature`, and `aFourthFeature`.
+
+The following three sections will illustrate how an orchestrating tool shouuld identify dependencies between Features, depending on [how the Feature is referenced](https://containers.dev/implementors/features/#referencing-a-feature).
+
+#### Identifying Dependencies - OCI Registry
+
+To speed up dependency resolution, Features [published to an OCI registry](https://containers.dev/implementors/features-distribution/#oci-registry) will have an annotation added to their manifest.  This annotation will be an escaped JSON object of the entire `dependsOn` object in the Feature's `devcontainer-feature.json`.  The orchestrating tool can use this annotation to resolve the Feature's dependencies without having to download and extract the Feature's tarball.
```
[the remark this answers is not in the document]
joshspicer: That's a great idea.  My only immediate concern was a size restriction, but I don't see anything [on the spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md) in that regard (and `devcontainer-feature.json` files are generally quite small and I don't foresee that changing.). I believe I would need to amend this to add `installsAfter` anyway (given the discussion you summarized below). 

I will update to serialize the entire `devcontainer-feature.json` onto the manifest 👍

nanis94:
main

proposals/feature-dependencies.md
nanis94: master

proposals/feature-dependencies.md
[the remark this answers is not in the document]
nanis94: main

nanis94 (commented):
master

nanis94 (commented):
main

nanis94 (commented):
joshspicer/dependsOn

nanis94 (commented):
joshspicer/dependsOn main

proposals/feature-dependencies.md
```diff
@@ -0,0 +1,206 @@
...
+
+#### (B2) Round-based sorting
+
+Perform a sort on the result of (B1) in rounds. This sort will rearrange and deduplicate Features, producing a sorted list of Features to install.  The sort will be performed as follows: 
+
+Start with all the elements from (B1) in a `worklist` and an empty list `installationOrder`.  While the `worklist` is not empty, iterate through each element in the `worklist` and check if all its dependencies (if any) are already members of `installationOrder`.  If the check is true, add it to an intermediate list `round` and remove it from the `worklist`.  If not, skip it.  Equality is determined in **Feature Equality**.
+
+Before committing each `round` to the `installationOrder`, sort all elements of round according to **Round Stable Sort***.
+
+Repeat for as many rounds as necessary until the worklist is empty.  If there is ever a round where no elements are added to `installationOrder`, the algorithm should terminate and return an error.  This indicates a circular dependency or other error in the dependency graph.
```
samruddhikhandale, line 134:
```suggestion
Repeat for as many rounds as necessary until the `worklist` is empty.  If there is ever a round where no elements are added to `installationOrder`, the algorithm should terminate and return an error.  This indicates a circular dependency or other error in the dependency graph.
```

proposals/feature-dependencies.md
```diff
@@ -0,0 +1,206 @@
...
+One implementation strategy involves assigning each node in the `worklist` a roundPriority integer, defaulting to 0.  If a Feature is indicated in `overrideFeatureInstallOrder`, the orchestrating tool should set the roundPriority of that Feature to  `length(overrideFeatureInstallOrder) - idx`, where `idx` is the zero-based index of the property in the array.  The orchestrating tool should then sort the `worklist` by roundPriority before committing each round's installation order, placing higher priority Features earlier in the `installationOrder`.
+
+## Additional Remarks
+
+### Feature authorship
+
+Features should be authored with the following considerations:
+
+- Features should be authored with the assumption that they will be installed in any order, so long as the dependencies are met at some point beforehand.
+- Since two Features with different options are considered different, a single Feature may be installed more than once.  Features should be idempotent.
```
samruddhikhandale, line 198: I am doubtful if the devcontainer/features can be installed multiple times. Curious, when do we update them? (before or after the CLI changes involving this spec)

proposals/feature-dependencies.md
```diff
@@ -0,0 +1,206 @@
...
+- Features that require updating shared state in the container (eg: updating the `$PATH`), should be aware that the same Feature may be run multiple times. Consider a method for previous runs of the Feature to communicate with future runs, updating the shared state in the intended way.
```
samruddhikhandale, line 199: 💡  Mostly, Features updates `PATH` var using `containerEnv`. I wonder if we should update CLI to handle this automatically.

proposals/feature-dependencies.md
```diff
@@ -0,0 +1,206 @@
...
+
+
+### Image Metadata
+
+Dependency resolution is only performed on initial container creation by reading the `features` object of `devcontainer.json` and resolving the Features in that point in time.  For subsequent creations from an image (or resumes of a dev container), the dependency tree is **not** re-calculated.  To re-calculate the dependency tree, the user must delete the image (or dev container) and recreate it from a `devcontainer.json`.
```
samruddhikhandale, line 204: > To re-calculate the dependency tree, the user must delete the image (or dev container) and recreate it from a `devcontainer.json`.

Would rebuilding update the dependency tree?

proposals/feature-dependencies.md
```diff
@@ -158,22 +158,35 @@ To prevent non-deterministic behavior, the algorithm will sort each **round** ac
...
 
-It is not recursive, and only influences the installation order of Features that are already set to be installed via a user-defined Feature.  It is not recursive, and does not influence the installation order of dependencies of the Feature that declares the `installsAfter` property.
+### overrideFeatureInstallOrder
+
+The `overrideFeatureInstallOrder` property of `devcontainer.json` is an array of Feature IDs that are to be installed in descending priority order as soon as its dependencies outlined above are installed.  
+
+This evaluation is performed after **Round Stable Sort** and may result in a different installation order within a round.
+
+This property must not influence the dependency relationship as defined by the dependency graph (see **(B1)**) and shall only be evaulated at the round-based sorting step (see **(B2)**).  Put another way, this property cannot "pull forward" a Feature until all of its dependencies (both soft and hard) have been installed.  After a Feature's dependencies have been installed in other rounds, this property should "pull forward" each Feature as early as possible (given the order of identifiers in the array).
```
chrmarti, line 183: Previously this could override the order given by soft dependencies. I think it is fine to change this since the main use case is to move features like `common-utils` which do not have any dependencies to the front of the install order. (Thinking out loud.)

proposals/feature-dependencies.md
joshspicer: Yes, recreating the container would require resolving and re-installing Features.

proposals/feature-dependencies.md
joshspicer: Without this property being included, the installation algorithm should behave as it did previously. Tactically, the Features in `devcontainers/features` will be updated and the major version incremented for them to take advantage of dependencies.  Other Features may choose as they please to adopt dependencies

proposals/feature-dependencies.md
```diff
@@ -116,22 +115,35 @@ Published Features referenced directly by [HTTPS URI pointing to the packaged ta
...
+
+The graph may be stored as an adjacency list with two kinds of edges (1) `dependsOn` edges or "hard dependencies" and (2) `installsAfter` edges or "soft dependencies".
 
-#### (B1) Building dependency graph
+#### **(B2) Assigning `roundPriority`**
 
-From the user-defined Features, the orchestrating tool will build a dependency graph.  The graph will be built by traversing the `dependsOn` property of each Feature and recursively resolving each Feature from the property.  An accumulator is maintained with each new Feature that has been discovered, as well as a pointer to its dependencies.  If the exact Feature (see **Feature Equality**) has already been added to the accumulator, it will not be added again.  The accumulator will be fed into (B2) after all the Feature tree has been resolved.
+Each node in the graph will originally be assigned a default `roundPriority` of 0.
```
samruddhikhandale, line 130:
```suggestion
Each node in the graph will originally be assigned a default `roundPriority` of 0 (**).  
```

proposals/feature-dependencies.md
```diff
@@ -116,22 +115,35 @@ Published Features referenced directly by [HTTPS URI pointing to the packaged ta
...
-#### (B1) Building dependency graph
+#### **(B2) Assigning `roundPriority`**
 
-From the user-defined Features, the orchestrating tool will build a dependency graph.  The graph will be built by traversing the `dependsOn` property of each Feature and recursively resolving each Feature from the property.  An accumulator is maintained with each new Feature that has been discovered, as well as a pointer to its dependencies.  If the exact Feature (see **Feature Equality**) has already been added to the accumulator, it will not be added again.  The accumulator will be fed into (B2) after all the Feature tree has been resolved.
+Each node in the graph will originally be assigned a default `roundPriority` of 0.  
 
+To influence installation order globally while still honoring the dependency graph of built in **(B1)**, `roundPriority` values may be tweaks for each Feature.  When each round is calculated in **(B3)**, only the Features equal to the max `roundPriority` of that set will be committed (the remaining will be uncommitted and reevaulated in subsequent rounds).
 
-#### (B2) Round-based sorting
+The `roundPriority` is set to a non-default value in the following instances:
```
samruddhikhandale, line 134:
```suggestion
(**) The `roundPriority` is set to a non-default value in the following instances:
```

proposals/feature-dependencies.md
samruddhikhandale: Suggestion: I believe it's not always true (see :L134)? Something like this might help avoid possible confusion.

samruddhikhandale (approved):
Looks great to me! 👏

chrmarti (changes requested):

proposals/feature-dependencies.md
```diff
@@ -116,22 +115,35 @@ Published Features referenced directly by [HTTPS URI pointing to the packaged ta
 
 ### (B) `dependsOn` install order algorithm
 
-The orchestrating tool is responsible for calculating a Featuure installation order (or error, if no valid installation order can be resolved). The set of Features to be installed is the union of user-defined Features and their dependencies.  The orchestrating tool will perform the following steps:
+The orchestrating tool is responsible for calculating a Feature installation order (or error, if no valid installation order can be resolved). The set of Features to be installed is the union of user-defined Features and their dependencies.  The orchestrating tool will perform the following steps:
+
+#### **(B1) Building dependency graph**
+
+From the user-defined Features, the orchestrating tool will build a dependency graph.  The graph will be built by traversing the `dependsOn` and `installsAfter` properties of each Feature.  The metadata for each dependency is then fetched and the node added as an edge to to the dependent Feature.  For `dependsOn` dependencies, the dependency will be fed back into the worklist to be recursively resolved.
```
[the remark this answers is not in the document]
joshspicer: Features with "soft dependencies" need their metadata "fetched" too, `installsAfter` properties are evaluated at the same time as the "hard dependencies"

samruddhikhandale (approved):
👏 👏

Issue #168, opened 2022-12-23
Docs says `initializeCommand` supports object, but not reflected in schema.json


sarisia:
Docs for devcontainer.json says `initializeCommand` supports object syntax, which will result in parallel execution, but spec schema.json doesn't reflect this.

docs:

https://github.com/devcontainers/spec/blob/b73bc6713b45e1f1bbbdd3a7d597f06faab41874/docs/specs/devcontainerjson-reference.md?plain=1#L70

https://github.com/devcontainers/spec/blob/b73bc6713b45e1f1bbbdd3a7d597f06faab41874/docs/specs/devcontainerjson-reference.md?plain=1#L108-L111

while schema.json doesn't allow `object` type:

https://github.com/devcontainers/spec/blob/b73bc6713b45e1f1bbbdd3a7d597f06faab41874/schemas/devContainer.base.schema.json#L230-L239

Chuxel:
I think the docs are wrong, correct @chrmarti ?

i-ky:
@sarisia, @Chuxel, @chrmarti, this can be closed now, I believe. I've just added `initializeCommand` to my configuration, object notation was in suggestions and seems to work as expected.

PR #237, opened 2023-05-09
add object-syntax for initializeCommand

joshspicer:
Reference Implementation: https://github.com/devcontainers/cli/pull/514

refs: https://github.com/microsoft/vscode-remote-release/issues/7765, https://github.com/devcontainers/spec/issues/168

Aligns the `initializeCommand` with the same syntax and behavior of the other lifecycle hooks.

### DIFF
--- a/schemas/devContainer.base.schema.json
+++ b/schemas/devContainer.base.schema.json
@@ -263,11 +263,21 @@
 				"initializeCommand": {
 					"type": [
 						"string",
-						"array"
+						"array",
+						"object"
 					],
-					"description": "A command to run locally before anything else. This command is run before \"onCreateCommand\". If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell.",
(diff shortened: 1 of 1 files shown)


schemas/devContainer.base.schema.json
```diff
@@ -263,11 +263,21 @@
 				"initializeCommand": {
 					"type": [
 						"string",
-						"array"
+						"array",
+						"object"
 					],
-					"description": "A command to run locally before anything else. This command is run before \"onCreateCommand\". If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell.",
+					"description": "A command to run locally (i.e Your host machine, cloud VM) before anything else. This command is run before \"onCreateCommand\". If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell. If this is an object, each provided command will be run in parallel.",
```
samruddhikhandale, line 269:
```suggestion
					"description": "A command to run locally (i.e your host machine, cloud VM) before anything else. This command is run before \"onCreateCommand\". If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell. If this is an object, each provided command will be run in parallel.",
```

samruddhikhandale (approved):
Thanks, left a minor comment!

chrmarti (approved):

PR #243, opened 2023-05-25
Set `source` property as not required for `Mount`

alexander-smolyakov:
**Description**:

The `source` attribute could be omitted for the `mount` command. In this case, Docker uses anonymous volumes. We update the devcontainer CLI to support such a case, and we need to update schemas to set the `source` property as optional.

_Changelog_:

- The `source` property was removed from `required` array;

**Related changes**:

- https://github.com/devcontainers/cli/pull/537

### DIFF
--- a/schemas/devContainer.base.schema.json
+++ b/schemas/devContainer.base.schema.json
@@ -688,7 +688,6 @@
 			},
 			"required": [
 				"type",
-				"source",
 				"target"
 			],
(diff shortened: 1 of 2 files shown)


chrmarti (approved):

avidal:
It looks like the referenced PR was merged and released already? Should this be merged now?

PR #260, opened 2023-06-20
fix: allow `$schema` property in `devcontainer.json`

### LINKED ISSUE #259: Permit `$schema` property in `devcontainer.json`
The current JSON schema definitions for `devcontainer.json` don't explicitly allow a `$schema` property, and `additionalProperties` is explicitly set to `false` at the top-level. This results in a warning of `Property $schema is not allowed.` when strictly evaluating a `devcontainer.json` against the schema.

A top-level property of `$schema` should be allowed with the following definition:

```json
"$schema": {
    "type": "string",
    "format": "uri"
}
```

This is the definition used throughout the JSON schema meta-schema.

JamieMagee:
Closes #259

### DIFF
--- a/schemas/devContainer.base.schema.json
+++ b/schemas/devContainer.base.schema.json
@@ -7,6 +7,11 @@
 		"devContainerCommon": {
 			"type": "object",
 			"properties": {
+				"$schema": {
+					"type": "string",
+					"format": "uri",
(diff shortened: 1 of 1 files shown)


samruddhikhandale (approved):
Thanks!

PR #277, opened 2023-07-14
Fix documentation markdown styling

l0rem1psum:
Found this issue while browsing the spec. Without proper styling, it is unable to be displayed in the browser.

This is what it currently looks like:

image


joshspicer (approved):
Thanks! I also opened https://github.com/devcontainers/devcontainers.github.io/pull/256 to update the website source (it's not directly shared from this repo)

PR #284, opened 2023-08-03
Make merge logic table less confusing

ronakj:
`x` is confusing in the supported fields entries because someone can assume that it means not supported.
`✓` is much less confusing.

### DIFF
--- a/docs/specs/devcontainer-reference.md
+++ b/docs/specs/devcontainer-reference.md
@@ -58,32 +58,32 @@ To apply the metadata together with a user's devcontainer.json at runtime the fo
 
 | Property | Type/Format | Merge Logic | devcontainer.json | devcontainer-feature.json |
 | -------- | ----------- | ----------- | :---------------: | :--------------: |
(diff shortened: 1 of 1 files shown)


bamurtaugh:
Thanks for taking the time to open this PR @ronakj!

We definitely want to make sure our references are as clear as possible. I suggest we keep this PR open to see if other folks from the community share any thoughts.

avidal:
So, all this changes is `x` to `✓`? If so, sounds good to me.

ronakj:
@bamurtaugh Created this PR in the other repo https://github.com/devcontainers/devcontainers.github.io/pull/286.

I also added `gpu` option for `hostRequirements` to this PR, as it was missing here but present in the other repo.

bamurtaugh (approved):
LGTM, thanks!

bamurtaugh:
@samruddhikhandale would you mind reviewing too? I think merge requires approval from someone in the maintainers org.

samruddhikhandale (approved):
✨

Issue #293, opened 2023-08-16
Consider using Bikeshed?


jcbhmr:
THIS IS HEAVILY OPINIONATED

from my background of reading and referencing specs like https://datatracker.ietf.org/doc/html/rfc2616 and https://html.spec.whatwg.org/multipage/workers.html and https://tc39.es/ecma262/multipage/global-object.html#sec-globalthis and https://wicg.github.io/webusb/ and https://drafts.csswg.org/selectors/#specificity-rules ... etc i have been conditioned to think of specifications as interconnected documents with lots of helpful links and terms and more in-page links and even more cross-spec links. I've found it a bit annoying having to open like 3 tabs to get a handle on how devcontainer.json interacts and combines with devcontainer-feature.json. 😅 and even then its a bunch of Ctrl+F-ing around the document to figure it out. i think it would be helpful if there was in-page links to other in-page terms.

in fact, i can see that there was at least some kind of effort to label terms with `**term** is a term` like text throughout the document:

```md
A **development container** is a container in which a user can develop an application.
```

i think that it may be worth at least considering (and maybe even exploring) the idea of using https://speced.github.io/bikeshed/ or even just improving the markdown to make it easier to read the specification.

for example...

```
A <dfn>development container</dfn> is a container in which a user can develop an application.
```

this AUTOMATICALLY creates a `#development-container` id-ed `<dfn>` and then later you can:

```
And then inside your [=development container=] you can use a [=dev container feature=] to ...
```
which autolinks to it.

bikeshed seems to make spec writing easier (since that's its entire purpose lol), but of course this could be done in plain markdown too.

on the other hand, i know there's the docker-related specs which are much different and look more like multipage booklets https://docs.docker.com/compose/compose-file/ which i don't have as much experience reading.

ps i think this would solve https://github.com/devcontainers/devcontainers.github.io/issues/10 https://speced.github.io/bikeshed/#id-gen and related autolinking confusion

bamurtaugh:
Thanks for opening @jcbhmr! I think this is a really interesting idea. I'll try to allot some time in the next few weeks to play around with Bikeshed. 

Either way, your points about consistency make a lot of sense, and it'd be valuable to set aside time to do a formatting pass (i.e. naming, capitalization, bolding).

For other folks who may be reviewing this - we definitely welcome PRs if you'd like to take a formatting pass or identify specific areas of inconsistency. 😄

bamurtaugh:
[...]

PR #322, opened 2023-10-20
Edit pass

bamurtaugh:
Edit pass similar to what I opened for containers.dev: https://github.com/devcontainers/devcontainers.github.io/pull/307

A few additional questions:
- Is there a reason we don't have Templates documentation in the `specs` folder? cc @samruddhikhandale 
- I think we could explore using Bikeshed as the user described in https://github.com/devcontainers/spec/issues/293#issuecomment-1752259665, though this would be a pretty major change to both the spec repo and containers.dev / its backing repo
     - I'm not 100% if it's worth the potentially great overhead (to begin with at least), since we're no longer at a point where we're making huge changes to the spec all the time
     - I'm open to other thoughts here


samruddhikhandale:
@bamurtaugh Feel free to move https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-templates.md and https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-templates-distribution.md to the spec folder.

bamurtaugh:
Done!

samruddhikhandale (approved):
✨

PR #324, opened 2023-10-27
Add build options

Roemer:
As in discussion https://github.com/devcontainers/spec/discussions/301, I would like to propose a way to add cli arguments to a `build` command, similar how `runArgs` add cli arguments to the `run` command.
This is needed for example for overriding the network eg. `network=host` or for adding additional hosts. There is also a discussion about it over at vscode-remote: https://github.com/microsoft/vscode-remote-release/issues/3545

Unfortunately, `build.args` is already used for Docker `ARG`uments so we need to use something else.
I am open for naming proposals.

Things I thought about are:
- build.buildArgs
- build.cliArgs
- build.cliOptions
- build.options
- build.commandArgs
- build.commandOptions

PR #328, opened 2023-10-31
Added build cli options proposal

Roemer:
Added a simple proposal for the spec change in https://github.com/devcontainers/spec/pull/324

### DIFF
--- a/proposals/build-cli-options.md
+++ b/proposals/build-cli-options.md
@@ -1,0 +1,50 @@
+# [Proposal] Allow additional cli options to pass to the build command
+
+Related to:
+- https://github.com/devcontainers/spec/discussions/301
+- https://github.com/devcontainers/spec/pull/324
+- https://github.com/microsoft/vscode-remote-release/issues/3545
+
+## Goal
+
+Extend the build section of the configuration with an option to pass additional arguments via cli.
+
+## Proposal
+
+Introduce the property `cliOptions` to the `build` property in the schema as type `Array (string)`. This field contains elements that are directly passed as cli-arguments to the `build` command. It is similar to `runArgs` property which contains cli-arguments passed to the `run` command.
+
+###  Execution
+
+When a dev container is built, all entries from the `cliOptions` are passed to the `build` command after all other options.
+
+## Examples
+
+### Add a host
+
+The following example adds a host-definition while the image is being built:
+
+```jsonc
+{
+  "build": {
+    "dockerfile": "Dockerfile",
+    "cliOptions": [
+      "--add-host=host.docker.internal:host-gateway"
+    ]
+  },
+}
+```
+
+### Use host network
+
+The following example allows to sue the host network while building an image:
+
+```jsonc
+{
+  "build": {
+    "dockerfile": "Dockerfile",
+    "cliOptions": [
+      "--network=host"
+    ]
+  },
+}
+```
--- a/proposals/build-options.md
+++ b/proposals/build-options.md
@@ -11,11 +11,11 @@ Extend the build section of the configuration with an option to pass additional 
 
 ## Proposal
 
-Introduce the property `cliOptions` to the `build` property in the schema as type `Array (string)`. This field contains elements that are directly passed as cli-arguments to the `build` command. It is similar to `runArgs` property which contains cli-arguments passed to the `run` command.
+Introduce the property `options` to the `build` property in the schema as type `Array (string)`. This field contains elements that are directly passed as cli-arguments to the `build` command. It is similar to `runArgs` property which contains cli-arguments passed to the `run` command.
 
 ###  Execution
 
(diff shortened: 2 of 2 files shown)


proposals/build-cli-options.md
```diff
@@ -0,0 +1,50 @@
...
+- https://github.com/devcontainers/spec/pull/324
+- https://github.com/microsoft/vscode-remote-release/issues/3545
+
+## Goal
+
+Extend the build section of the configuration with an option to pass additional arguments via cli.
+
+## Proposal
+
+Introduce the property `cliOptions` to the `buildOptions` in the schema. It is very similar to `runArgs`, which contains cli-arguments passed to the `run` command, the `cliOptions` passes cli-arguments to the `build` command.
```
samruddhikhandale, line 14: I wonder if calling the new property as `options` makes sense. (eg `build.options`)

// cc @bamurtaugh @craiglpeters @Chuxel for product specific insights. Also, there are many other helpful suggestions from the author in the https://github.com/devcontainers/spec/pull/324 description. 

> build.buildArgs
build.cliArgs
build.cliOptions
build.options
build.commandArgs
build.commandOptions

proposals/build-cli-options.md
```diff
@@ -0,0 +1,50 @@
...
+
```
samruddhikhandale, line 15: Can we also add a line in the proposal stating that the new property will be of type `Array (string)`

samruddhikhandale (commented):
Thank you for taking the time in contributing this proposal, we appreciate it 👏

proposals/build-cli-options.md
Roemer: I am open for the naming. `options` seems a bit too generic for me. In the end, the field represents cli arguments passed to the command.

proposals/build-cli-options.md
Roemer: I adjusted the text a bit and also mention the type.

proposals/build-cli-options.md
```diff
@@ -0,0 +1,50 @@
...
+- https://github.com/devcontainers/spec/pull/324
+- https://github.com/microsoft/vscode-remote-release/issues/3545
+
+## Goal
+
+Extend the build section of the configuration with an option to pass additional arguments via cli.
+
+## Proposal
+
+Introduce the property `cliOptions` to the `buildOptions` in the schema as type `Array (string)`. This field contains elements that are directly passed as cli-arguments to the `build` command. It is similar to `runArgs` property which contains cli-arguments passed to the `run` command.
```
chrmarti, line 14: Should this be "`cliArgs` to the `build` property" to match existing names?

proposals/build-cli-options.md
```diff
@@ -0,0 +1,50 @@
...
+
+###  Execution
+
+When a dev container is built, all entries from the `cliOptions` are passed to the `build` command at the end.
```
chrmarti, line 18: Change "at the end" to "after all other options" since the context path will still go last?

chrmarti (changes requested):
Thanks for the proposal! Left a few questions.

proposals/build-cli-options.md
Roemer: Adjusted it, yes it's the `build` property (the type in the schema is `buildOptions`).

proposals/build-cli-options.md
Roemer: Good catch. True of course, adjusted it.

proposals/build-cli-options.md
bamurtaugh: Thanks for providing all the great suggestions! I lean toward `build.options` too - it feels the cleanest and most concise. While other areas may also have things called "options," I don't anticipate us wanting to refer to other things as `build.options`.

Roemer:
I just stumpled uppon problems with features as well. Should this proposal also add a point that those `options` should also be added when adding features?

proposals/build-cli-options.md
Roemer: I renamed it to `options`.

chrmarti:
The image is built in a single `docker build` run, adding build args to features might allow for conflicting arguments.

chrmarti (approved):
LGTM!

proposals/build-options.md
```diff
@@ -0,0 +1,50 @@
...
+    "options": [
+      "--add-host=host.docker.internal:host-gateway"
+    ]
+  },
+}
+```
+
+### Use host network
+
+The following example allows to sue the host network while building an image:
```
eljog, line 39: typo?
```suggestion
The following example allows to use the host network while building an image:
```

eljog (approved):
⚡

113ll3 (commented):
Looks like a buzz

PR #348, opened 2023-12-04
Doc update for life cycle hook features

dfberry:


docs/specs/devcontainer-features.md
```diff
@@ -56,13 +56,13 @@ All properties are optional **except for `id`, `version`, and `name`**.
...
-| :--- | :--- |
-| `onCreateCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties)|
-| `updateContentCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties)|
-| `postCreateCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties)|
-| `postStartCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties) |
-| `postAttachCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties) |
+| Property | Type|Description | 
+| :--- | :--- | :-- |
+| `onCreateCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties)|First time inside the container immediately after it has started. |
+| `updateContentCommand` | [string, array, object](/docs/specs/devcontainerjson-reference.md#formatting-string-vs-array-properties)| First time inside the container after `onCreateCommand` whenever new content is available in the source tree during the creation process.|
```
eljog, line 62: @joshspicer do you think we should use the wordings consistent to the documentation [here](https://containers.dev/implementors/json_reference/#lifecycle-scripts)

eljog (commented):
Thank you for your contribution
I have added a comment for a peer review

docs/specs/devcontainer-features.md
joshspicer: Would a link back to https://containers.dev/implementors/json_reference/#lifecycle-scripts on this page be sufficient clarity, @dfberry ? 

I think reducing the amount of duplicated descriptions would be helpful to prevent information getting stale.

Regardless I agree on the consistency

dfberry: There is a link back, it below but not super obvious. I tried to be very respectful of the text in the other doc. I didn't know these other life cycle hooks where available because dev containers in vscode really only shows 1 by default. I was trying to make it more obvious but perhaps this is the wrong platform for that? Not sure. @seesharprun @bamurtaugh Thoughts?

bamurtaugh: Thanks for opening this @dfberry. I love the goal @joshspicer mentions to avoid duplicated definitions and include links as possible, especially as we have content here and in containers.dev. 

@dfberry maybe if the link wasn't apparent, another link could be added on top of the table, and/or links could be added to each entry in the table?

And speaking of containers.dev, for any change we pursue in this repo, could you also open a PR in the containers.dev repo when you get a chance: https://github.com/devcontainers/devcontainers.github.io/blob/gh-pages/_implementors/features.md? Thank you!

dfberry: @bamurtaugh Let me step back and walk through how I got here because I definitely think a link is missing from the devcontainer created by VSCode to this understanding of the different lifecycle events which @seesharprun alerted me to. Once I got to this documentation for lifecycle properties, I had to figure out which lifecycle event was best suited to the typical usecases for scripts in the dev container. 

Issues are:
1) from devcontainers, how do I know and understand their are more lifecycle hooks than the 1 presented in the devcontainer
2) once I get to the documentation, how do I understand use case for each event

bamurtaugh: Thanks for clarifying. It sounds like an alternative option might be adding a link to these docs via a comment in the dev container Template or image, i.e. above this line https://github.com/devcontainers/templates/blob/main/src/cpp/.devcontainer/devcontainer.json#L16. Does that sound like it would've helped @dfberry?

RatStar811 (approved):

RatStar811 (approved):

Issue #349, opened 2023-12-12
JSON schema for dev container templates


benblank:
The spec defines dev container templates, but I was unable to find a corresponding JSON schema. The provided schemas for dev containers and dev container features are quite handy when creating them; I believe having a schema for templates as well would be a nice quality-of-life improvement for template authors.

PR #350, opened 2023-12-14
Add JSON schema for Development Container Template Metadata

benblank:
I've made an attempt at creating the JSON schema for Template metadata which I requested in #349.

Some notes:

* It is patterned after `devContainerFeature.schema.json`, on the theory that Template metadata has more in common with Feature metadata than it does Development Container metadata.
* The spec doesn't indicate which properties are required, so:
    * For Templates, between how they are described to be used and the fact that `id`, `name`, and `version` are required for Feature metadata, I've assumed those same three properties are required for Template metadata.
    * For options, based on how they are described to be used, I've assumed that all properties are required (with the exception of `enum` and `proposals`; see the points below).
* Though the spec indicates that an option's `default` property may only be a string, I've assumed that it should instead be a boolean when the `type` property is `"boolean"`.
* With regards to the `enum` and `proposals` properties of options:
    * I've assumed that neither `enum` nor `proposals` are relevant to boolean options, so have forbidden them.
    * I've assumed that `enum` cannot be an empty array, on the theory that because free-form values are not allowed, it would make it impossible to set a valid value for that option.
        * I have *not* made the same assumption regarding `proposals`, as free-form values are allowed.
    * As a result of the preceding point, if `enum` is not present on an option, `proposals` **must** be, even if it is an empty array.
* I haven't placed any restrictions (e.g. via patterns) on Template IDs, Template versions, or option IDs, as the Feature schema also doesn't. However, these could easily be added.

Apologies for the verbose description; I'm only getting started with Development Containers and want to make sure that any mistakes I've made are easy to spot. 😅

### DIFF
--- a/schemas/devContainerTemplate.schema.json
+++ b/schemas/devContainerTemplate.schema.json
@@ -1,0 +1,182 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Development Container Template Metadata",
+  "description": "Development Container Template Metadata (devcontainer-template.json). See https://containers.dev/implementors/templates/ for more information.",
+  "definitions": {
+    "Template": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "$schema": {
+          "description": "The JSON schema of the `devcontainer-template.json` file.",
+          "type": "string",
+          "format": "uri"
+        },
+        "description": {
+          "description": "Description of the Template.",
+          "type": "string"
+        },
+        "documentationURL": {
+          "description": "Url that points to the documentation of the Template.",
+          "type": "string",
+          "format": "uri"
+        },
+        "id": {
+          "description": "ID of the Template. The id should be unique in the context of the repository/published package where the Template exists and must match the name of the directory where the devcontainer-template.json resides.",
+          "type": "string"
+        },
+        "keywords": {
+          "description": "List of strings relevant to a user that would search for this Template.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "licenseURL": {
+          "description": "Url that points to the license of the Template.",
+          "type": "string",
+          "format": "uri"
+        },
+        "name": {
+          "description": "Name of the Template.",
+          "type": "string"
+        },
+        "options": {
+          "description": "A map of options that the supporting tools should use to populate different configuration options for the Template.",
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/Option"
+          }
+        },
+        "platforms": {
+          "description": "Languages and platforms supported by the Template.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "publisher": {
+          "description": "Name of the publisher/maintainer of the Template.",
+          "type": "string"
+        },
+        "version": {
+          "description": "The semantic version of the Template.",
+          "type": "string"
+        }
+      },
+      "required": [
+        "id",
+        "name",
+        "version"
+      ]
+    },
+    "Option": {
+      "description": "An option that the supporting tools should use to populate different configuration options for the Template.",
+      "type": "object",
+      "properties": {
+        "default": {
+          "description": "Default value for the option.",
+          "type": [
+            "boolean",
+            "string"
+          ]
+        },
+        "description": {
+          "description": "Description for the option.",
+          "type": "string"
+        },
+        "type": {
+          "description": "Type of the option. Valid types are currently: boolean, string",
+          "type": "string",
+          "enum": [
+            "boolean",
+            "string"
+          ]
+        }
+      },
+      "required": [
+        "default",
+        "description",
+        "type"
+      ],
+      "oneOf": [
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "default": {
+              "type": "boolean"
+            },
+            "description": true,
+            "type": {
+              "type": "string",
+              "const": "boolean"
+            }
+          },
+          "required": [
+            "default",
+            "type"
+          ]
+        },
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "default": {
+              "type": "string"
+            },
+            "description": true,
+            "enum": true,
+            "proposals": true,
+            "type": {
+              "type": "string",
+              "const": "string"
+            }
+          },
+          "required": [
+            "default",
+            "type"
+          ],
+          "oneOf": [
+            {
+              "type": "object",
+              "properties": {
+                "enum": {
+                  "description": "A strict list of allowed string values. Free-form values are not allowed. Omit when using optionId.proposals.",
+                  "type": "array",
+                  "minItems": 1,
+                  "items": {
+                    "type": "string"
+                  }
+                }
+              },
+              "required": [
+                "enum"
+              ]
+            },
+            {
+              "type": "object",
+              "properties": {
+                "proposals": {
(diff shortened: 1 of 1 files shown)


benblank:

schemas/devContainerTemplate.schema.json
TateGunning, line 1: Thankful, please advise

TateGunning (approved):
Looks fine :)

schemas/devContainerTemplate.schema.json
```diff
@@ -0,0 +1,182 @@
...
+            "string"
+          ]
+        }
+      },
+      "required": [
+        "default",
+        "description",
+        "type"
+      ],
+      "oneOf": [
```
chrmarti, line 102: devContainerFeature.schema.json uses `anyOf` here and has a few more descriptions filled in. Maybe copy from that?

schemas/devContainerTemplate.schema.json
```diff
@@ -0,0 +1,182 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Development Container Template Metadata",
+  "description": "Development Container Template Metadata (devcontainer-template.json). See https://containers.dev/implementors/templates/ for more information.",
+  "definitions": {
+    "Template": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "$schema": {
```
chrmarti, line 10: This would also make sense in the features schema.

chrmarti (changes requested):
Thanks for the PR! Looks great overall, I have left a few comments for consideration / discussion.

We should then also update https://github.com/devcontainers/action to use this schema similarly to how it uses the schema for features.

We could add the features and templates schemas to VS Code like we do for the main schema here: https://github.com/microsoft/vscode/blob/75b1639461b53dc51380a61532aabdd502c4c718/extensions/configuration-editing/package.json#L142

schemas/devContainerTemplate.schema.json
```diff
@@ -0,0 +1,182 @@
...
+          "type": "string"
+        },
+        "version": {
+          "description": "The semantic version of the Template.",
+          "type": "string"
+        }
+      },
+      "required": [
+        "id",
+        "name",
```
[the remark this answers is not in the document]
benblank: Hmm. It isn't marked as required in `devContainerFeature.schema.json` now that I look at it, but it is on [the implementors page](https://containers.dev/implementors/features/#devcontainer-feature-json-properties). I'm not yet familiar enough with the spec to know which is accurate, but I suppose the other should be updated to match. 🙂

schemas/devContainerTemplate.schema.json
benblank: I'll update the `oneOf` to `anyOf` to match the Feature schema and do another pass on the descriptions; I didn't think I'd missed any. 🙁

schemas/devContainerTemplate.schema.json
benblank: Will do.

schemas/devContainerTemplate.schema.json
benblank: Okay, I *completely* misunderstood what you were referring to here. 😅

I used `oneOf` here because it more clearly communicates that *exactly one* of the subschemas (boolean option, string option) must apply. In this case, they're incompatible (`"const": "boolean"` vs. `"const": "string"`), but even if they weren't, it wouldn't be reasonable for an option to be considered both a boolean and a string.

Similarly, the placement of the descriptions are deliberate, to avoid redundancy. The descriptions for the properties which are present regardless of the option's type (`default`, `description`, and `type`) are present at the top level so that those properties don't need a subschema with any more detail than is particular to the `oneOf` variant selected; you can see this most easily with the `description` property, which has the subschema of just `true` for both boolean options and string options.

Admittedly, I've only tested this with VS Code, but it behaves as I would expect it to: the descriptions for all properties are available regardless of which `oneOf` variant has been matched, or even if neither has yet matched.

"default" property description (defined in base option schema), no option type
*"default" property description (defined in base option schema), no option type*

"default" property description (defined in base option schema), option has "string" type
*"default" property description (defined in base option schema), option has "string" type*

"proposals" property description (defined in string option subschema), no option type
*"proposals" property description (defined in string option subschema), no option type*

"proposals" property description (defined in string option subschema), option has "string" type
*"proposals" property description (defined in string option subschema), option has "string" type*

schemas/devContainerTemplate.schema.json
benblank: I just pushed a new commit to the PR which adds this.

schemas/devContainerTemplate.schema.json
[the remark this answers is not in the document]
chrmarti: @joshspicer Should we make `name` required in the features schema? Would a feature work if it had no name currently? If not, we would know that making it required in the schema won't break anyone. :)

schemas/devContainerTemplate.schema.json
chrmarti: I see, it should be `oneOf` in the features schema then, could you update it there? Thanks!

Replicating the descriptions might still make sense for other tools and helps when someone reads the schema themselves. 🤔

benblank: No worries, I've made that change to the feature schema. While I was in there, I noticed that "proposals" isn't required when "enum" is absent and have made the same change to the template schema. (Thinking back, I don't know why I made it required in the first place.)

I'm a bit torn over the descriptions as well. On the one hand, schemas are primarily intended for machine consumption rather than being read directly by humans… but on the other hand, a lot of us read them anyway! As far as other consumers go, my experience with schema tools other than VS Code is pretty much limited to validation libraries, which I don't believe I've seen use descriptions from the schema.

Ultimately, I came down on the side of "redundancy is bad" simply because future updates can't cause duplicate descriptions to drift out of sync if there aren't duplicate descriptions in the first place. If there's a stronger argument for duplication that just hasn't occurred to me, I'm happy to make the change!

marcoantonio-af (approved):

schemas/devContainerTemplate.schema.json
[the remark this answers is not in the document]
joshspicer: In the tooling `name` is not required.  For example when we build the READMEs in `devcontainers/action` it will use `id` if `name` is not available.

That said, we should probably just make it required since it's useful as a display name in our tooling

One thing to note - if we add `name` as required in the schema for Features, this [validation](https://github.com/devcontainers/action/blob/main/src/utils.ts#L119) code when publishing with the `devcontainers/action` will fail if the Feature is missing `name`.

113ll3:
I'm going to be honest I am researching who or what and why I am getting
repositories i don't have any idea how and im experiencing lotsbof abnormal
iterference on my and. Droid devices so I have no clue I need assistance or
help to identify why my devices have like gecko HTML

schemas/devContainerTemplate.schema.json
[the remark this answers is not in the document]
chrmarti: It seems safer to update the spec to make it optional then. We could break existing features as you point out.

One advantage of not making it required is that it is easier to write one-off features for prototyping or testing. We would still want features going into the index to have good names for the UI. 🤷‍♂️

Xcopz1 (approved):

schemas/devContainerFeature.schema.json
```diff
@@ -211,7 +216,7 @@
       "type": "object"
     },
     "FeatureOption": {
-      "anyOf": [
```
willhao0430, line 214: >

willhao0430 (approved):

Qaisfaris (approved):

PR #381, opened 2024-01-11
Fix type error in spec example (updated :))

trukhinyuri:
Fix type error in spec example: overrideFeatureInstallorder -> overrideFeatureInstallOrder


joshspicer (approved):
Thanks!

PR #24, opened 2022-03-29
Specify jsonc 

bamurtaugh:
Specify devcontainer.json as jsonc.

PR #458, opened 2024-04-20
fix: allow trialing comma in JSON schema

popsu:
[Sorry :pray: if this is not the right way to do this, I just got here when I was looking how to silence 'Trailing comma' warnings in VS Code when editing a `devcontainer.json` file]

Currently a `devcontainer.json` with trailing commas works.

Either the schema or implementation is wrong, since they don't match. Changing the schema is probably easier, since changing the implementation would probably break many devcontainers.

### DIFF
--- a/schemas/devContainer.base.schema.json
+++ b/schemas/devContainer.base.schema.json
@@ -2,7 +2,7 @@
 	"$schema": "https://json-schema.org/draft/2019-09/schema",
 	"description": "Defines a dev container",
 	"allowComments": true,
-	"allowTrailingCommas": false,
+	"allowTrailingCommas": true,
 	"definitions": {
 		"devContainerCommon": {
 			"type": "object",


samruddhikhandale (approved):
Looks like an acceptable change to me!

chrmarti (changes requested):
We previously disallowed it: https://github.com/devcontainers/spec/pull/24#issuecomment-1091846135

popsu:
> We previously disallowed it: [#24 (comment)](https://github.com/devcontainers/spec/pull/24#issuecomment-1091846135)

The implementation allows it tho. My issue is that the implementation doesn't match spec and thus I get some warnings about invalid files. Sure it's not that big of an issue, but maybe other peopled had the same small issue :sweat_smile:

But I guess I'll just close this since it's not that big of a deal in the end, and seems like it's not that clear-cut how all this 'JSONC' even works and how are different editors using it.

schemas/devContainer.base.schema.json
```diff
@@ -2,7 +2,7 @@
 	"$schema": "https://json-schema.org/draft/2019-09/schema",
 	"description": "Defines a dev container",
 	"allowComments": true,
-	"allowTrailingCommas": false,
+	"allowTrailingCommas": true,
```
Badboyboom, line 5: popsu:patch-1

PR #484, opened 2024-08-01
Proposal: Template `optionalPaths` property

joshspicer:
Related contexts: 
- https://github.com/microsoft/vscode-remote-release/issues/10095#issuecomment-2260982054 
- https://github.com/devcontainers/templates/issues/274

Reference implementation change: 
- https://github.com/devcontainers/cli/pull/868

Related changes:
- https://github.com/devcontainers/cli/pull/865
- https://github.com/devcontainers/cli/pull/866

### DIFF
--- a/docs/specs/devcontainer-templates.md
+++ b/docs/specs/devcontainer-templates.md
@@ -71,8 +71,7 @@ The `options` property contains a map of option IDs and their related configurat
 
 Before applying a Template, tooling must inspect the `optionalPaths` property of a Template and prompt the user on whether each file or folder should be included in the resulting output workspace folder.  A path is relative to the root of the Template source directory.
 
-- For a single file, provide the full relative path without any path delimiters.  
-
+- For a single file, provide the full relative path (without any leading or trailing path delimiters).  
 - For a directory, provide the full relative path with a trailing slash and asterisk (`/*`) appended to the path.  The directory and its children will be recursively ignored.
 
 Examples are shown below:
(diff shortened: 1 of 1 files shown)


docs/specs/devcontainer-templates.md
```diff
@@ -66,6 +67,26 @@ The `options` property contains a map of option IDs and their related configurat
...
+```json
+{
+    "id": "cpp",
+    "version": "3.0.0",
+    "name": "C++",
+    "description": "Develop C++ applications",
+    "optionalPaths": [
+         "GETTING-STARTED.md",
+         "example-project-1/MyProject.csproj",
+         ".github/"
```
samruddhikhandale, line 85: I wonder if folks might forget to add the `/` suffix to denote a folder. Should we allow `/*` instead, which seems more familiar (at least to me)? However, I am definitely open to your thoughts on this 👀

joshspicer: Thanks for the feedback! 

> I wonder if folks might forget to add the / suffix to denote a folder and the CLI might incorrectly work?

In my tagged implementation omitting the `/` wouldn't cause any of its child files to be included, so it would work as intended

> Should we allow /* instead, which seems more familiar 

Sounds reasonable to me.  I don't think we have such a precedent in the dev containers spec for how this should look.  

Though, perhaps i'm overthinking it, as all of `.github`, `.github/`, and `.github/*` would get copied with the `cp -r` command. 

Another way this could be phrased is `optionalPathPrefixes`, where anything under the prefix gets copied (that's essentially what is happening here but always with an implicit `/*` at the end matching anything that shares that path prefix.

samruddhikhandale (approved):
Wonderful ⚡

Kutter07 (approved):

PR #508, opened 2024-10-16
Create devcontainer.json

mustafacco7:


bamurtaugh:
Thanks for your contribution! I don't think this is a necessary change, but please let us know if you have other context.

PR #557, opened 2025-03-03
Create devcontainer.json

haticecan37:
`gh repo clone devcontainers/spec `


bamurtaugh:
If there's a specific reason you think this change would be helpful, please let us know. Thank you!

PR #568, opened 2025-04-06
fix: incorrect reference to `volumes` attribute

MasonM:
`mounts` is not a valid attribute in Docker Compose files. I'm assuming this is a typo for [`volumes`](https://docs.docker.com/reference/compose-file/services/#volumes).


bamurtaugh (approved):
Thank you! cc @eljog for your review

DeezDevelopes (commented):
thanks cutie

DeezDevelopes (approved):
thanks cutie

PR #571, opened 2025-04-15
Add proposal and update schema

scriptjs:
The following provides a clear proposal for the inclusion of the array and object types for Feature options. The updated schema validate JSON Schema 7 via https://www.jsonschemavalidator.net/ and was tested against the samples provided in the proposal.

### DIFF
--- a/proposals/add-array-object-feature-option-types.md
+++ b/proposals/add-array-object-feature-option-types.md
@@ -1,0 +1,170 @@
+# Proposal: Add `array` and `object` Types to Feature Options Schema
+
+## Problem Statement
+
+The current Dev Container Feature Options Schema lacks support for `array` and `object` types. This limitation hinders the ability to configure features that require lists of values (e.g., dependencies, ports) or key-value mappings (e.g., environment variables, custom settings). Addressing this limitation will enhance the flexibility and usability of the schema.
+
+This proposal aims to address [Issue #567](https://github.com/devcontainers/spec/issues/567).
+
+## Proposed Solution
+
+Extend the Dev Container Feature Options Schema by introducing `array` and `object` as valid types for feature options. These new types will enable developers to create more complex and customizable features.
+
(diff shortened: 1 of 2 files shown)


scriptjs:

proposals/add-array-object-feature-option-types.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Problem Statement
+
+The current Dev Container Feature Options Schema lacks support for `array` and `object` types. This limitation hinders the ability to configure features that require lists of values (e.g., dependencies, ports) or key-value mappings (e.g., environment variables, custom settings). Addressing this limitation will enhance the flexibility and usability of the schema.
+
+This proposal aims to address [Issue #567](https://github.com/devcontainers/spec/issues/567).
+
+## Proposed Solution
+
+Extend the Dev Container Feature Options Schema by introducing `array` and `object` as valid types for feature options. These new types will enable developers to create more complex and customizable features.
```
chrmarti, line 11: What types of values would be supported? Maybe limiting it to strings and booleans would keep it simple and aligned with the existing types.

proposals/add-array-object-feature-option-types.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+**Installation Script (`example-feature/install.sh`):**
+
+```bash
+#!/bin/bash
+
+# Install dependencies
+echo "Installing dependencies..."
+for package in "${DEPENDENCIES[@]}"; do
```
chrmarti, line 65: This seems to be Bash specific syntax. How would this work with the Bourne shell `/bin/sh`? The Bourne shell is used for the install scripts to allow for compatibility with distributions not including Bash by default like Alpine Linux.

chrmarti (changes requested):
Thanks for the proposal! Left a few questions.

PR #674, opened 2025-12-14
Proposal: Add localFeaturesRoot for flexible local feature paths

### LINKED ISSUE #671: Consider making it easier to reuse local features
# Background

I'm looking at dev containers more and I've been aiming to have a single repository to manage both my own:

* features
* dev container configurations

The current spec makes *reusing* local features between dev container configurations trickier than perhaps it could be.

I have managed to achieve this in some form, but it involved both considerable workarounds and considerable trial and error to see what would work. 

## Benefits of local features

Why are local features useful?

For individual developers trying to manage a collection of their own common configurations for reuse, local features running directly from files are nicely lightweight. Local features let you:

* avoid building, publishing and managing full-blown feature packages
* change the configuration and have those changes take direct effect on next dev container build, without any further steps
* keep features private easily, without worrying about usage quotas/hosting costs (example: GitHub charge for private package use)[^1]

# The current restrictions

There are [restrictions in the specification](https://containers.dev/implementors/features-distribution/#addendum-locally-referenced) on where features are stored: they must be in the workspace root.[^2]

Clunky workarounds are needed to be able to have a working local feature stored elsewhere, more "centrally", and not only in a specific configuration for a specific dev container configuration.

### Local features do not support VS Code's alternative dev container configuration paths

* VS Code offers a way to [specify alternative paths for dev container configurations](https://code.visualstudio.com/docs/devcontainers/create-dev-container#_alternative-repository-configuration-folders), other than the original source repository folder itself.
* When you open the *source repository folder*, the dev container configuration stored elsewhere is then used.
* *This is good: it gets part way to managing dev container setups in a completely separate path, which could also be a version controlled repository itself.*
* Unfortunately, move your entire functioning dev container configuration from the repository to an alternative configuration folder doesn't work. The dev container will start to build. But the features are no longer found within the `.devcontainer` folder *in the workspace root*. Building the dev container then fails.
* (Maybe there are workarounds of linking the features back in to the original source repository here? But this is particularly messy.)
* I described a workaround in a [vscode-remote-release issue](https://github.com/microsoft/vscode-remote-release/issues/11356) where you instead open the *dev container configuration folder*, and specify the intended repository folder as a mount. This finds the features as part of the workspace root then, and builds correctly.

### Local features still can't be easily reused, even then

* Even with the workaround above, this still does not give me a way to easily create the same feature once, and then re-use it.
* Further workarounds to have one canonical "source" feature, but get the feature files in place inside the dev container configuration are needed, such as:
  * hardlinking the shared feature files into each dev container configuration where they are used (symlinks are insufficient)
  * adding the shared features to a Git submodule containing all features, pushing that, and cloning it within each dev container configuration for use

All this just about _works_ 🔧, but has a lot more friction in doing so.

[^1]: Yes, I could publish publicly, but these features are for my particular needs and probably of very little value to anyone else
[^2]: I think I understand where this comes from; it's focused on the use case of a specific configuration stored within a specific repository.

tennisleng:
## Summary

This proposal addresses **[Issue #671](https://github.com/devcontainers/spec/issues/671)** - making it easier to reuse local Features in dev containers.

## Problem

Users who want to manage a collection of their own dev container configurations and Features in a single repository face significant friction. The current specification requires local Features to be stored within the `.devcontainer/` folder at the project workspace folder root, which makes reusing Features across different dev container configurations difficult.

## Solution

This PR introduces a new `localFeaturesRoot` property for `devcontainer.json` that allows users to specify a custom base directory for resolving local Feature paths.

### Property Definition

|----------|------|-------------|
| `localFeaturesRoot` | string | Path (relative to the folder containing `devcontainer.json`) that serves as the base for resolving local Feature paths. Defaults to `.` (current behavior). |

### Example Usage

```jsonc
{
    "name": "My Project",
    "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
    "localFeaturesRoot": "../../shared-features",
    "features": {
        "./my-feature": {}  // Resolves to shared-features/my-feature
    }
}
```

## Changes Made

1. **proposals/flexible-local-feature-paths.md** - New proposal document with full specification
2. **docs/specs/devcontainerjson-reference.md** - Added `localFeaturesRoot` property to the reference
3. **docs/specs/devcontainer-features-distribution.md** - Updated "Locally referenced Features" section with new property documentation and examples
4. **schemas/devContainer.base.schema.json** - Added `localFeaturesRoot` property to the JSON schema

## Backward Compatibility

This change is fully backward compatible:
- The `localFeaturesRoot` property is optional
- When not specified, behavior is identical to current spec
- Existing configurations continue to work without modification

## Implementation Notes

This is a **specification-only** proposal. The actual implementation would need to be done in the [devcontainers/cli](https://github.com/devcontainers/cli) repository.

---

Fixes #671


tennisleng:

PR #760, opened 2026-08-16
Add $schema property for devcontainer-feature schema

rob9315:
mirrored how the schema for `devcontainer.json` does this: https://github.com/devcontainers/spec/blob/c95ffeed1d059abfe9ffbe79762dc2fa4e7c2421/schemas/devContainer.base.schema.json#L10-L14

to avoid silly warnings like these:
image


TheVolpeDev (approved):
LGTM;

